PR #69, opened 2022-08-05
Changed behaviour for get cloudevent

tba76:
Update db-call to run async.
See documentation:
https://www.npgsql.org/doc/basic-usage.html


acn-sbuad (approved):
🎈

PR #74, opened 2022-08-10
Add index for id column in events.events table

SandGrainOne:
## Description
Add a new index for the id column in order to improve the performance of queries using the id as a filter.


tba76 (approved):
🍺

Issue #129, opened 2022-09-20
Subscription duplicate check


SandGrainOne:
### Description

The current logic will allow the same consumer to create multiple identical subscriptions. The Subscription logic should be able to stop this from happening.

Same subscription is identified when all properties are the same as an existing subscription. Excluding ID and alternative subject.

If a duplicate is found:
- The endpoint returns 201 with the original subscription.
- Trigger subscription validation. (if not already valid)

### Tasks

- [x] Create a new function that can look for existing subscription based on properties from a new one
- [x] Change logic in service to first look for existing subscription. Return existing if found instead of saving new.
- [x] Expand function return columns to include validation.

### Acceptance Criterias

- [x] The POST endpoint for subscription returns the ID of an existing subscription if it matches the new properties
- [x] The re-registration of a subsription triggers validation if the previous validation status was invalid.
- [x] The returned subscription includes validation status (true/false)

PR #145, opened 2022-09-28
Feature/129 subscription duplicate check

SandGrainOne:
## Description
Adding a duplicate check to the create subscription logic. This is to silently stop a client from creating identical duplicate subscriptions.

## Related Issue(s)
- #129 

## Verification
- [x] All tests run green

## Documentation


src/Events/Models/Subscription.cs
```diff
@@ -54,6 +54,11 @@ public class Subscription
         /// </summary>
         public DateTime Created { get; set; }
 
+        /// <summary>
```
vgr, line 57: What are the criterias for ok? - That the endpoint has validated the PUT request?

vgr (approved):
Looks good to me.

src/Events/Models/Subscription.cs
SandGrainOne: Yes, the endpoint replied with 200 OK.

PR #147, opened 2022-09-30
roll back Microsoft.IdentityModel.Tokens to 6.23.0

acn-sbuad:
## Description
V 6.23.1 of the package causes function to fail during token generation. Must investigate cause of error and request fix from microsoft before patching this library. 

## Related Issue(s)
- #{issue number}
N/A

## Verification


vgr (approved):
LGTM

Issue #117, opened 2022-09-16
Refactoring: Moving business logic down to service


SandGrainOne:
### Description

There are currently some logic in controllers that we need to move down to the services/businesslogic.

Example:
- AppEventsController.GetForOrg should not do the party lookup. Move to service.

### Considerations

In SubscriptionsController, which way should we go to communicate errors from service to controller? 
- Throw exceptions? 
- Return a problem object with status code and message? 
- Other? 

### Tasks

- [x] AppEventsController
  - [x] Unit tests
  - [x] QA 
- [x] PushEventsController
  - [x] Unit tests
  - [x] QA 
- [x] SubscriptionsController
  - [x] Unit tests
  - [x] QA 
- [x] Regression test
- [ ] Update handbook on docs with how we transfer "status results" from service to controller.
### Acceptance Criterias

- [x] Controllers have dependencies only to 1 or 2 services that have the business logic.

PR #148, opened 2022-09-30
added missig check in appeventsservice

acn-sbuad:
## Related Issue(s)
- #117

## Verification
- [x] All tests run green

### DIFF
--- a/src/Events/Services/AppEventsService.cs
+++ b/src/Events/Services/AppEventsService.cs
@@ -2,7 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Text.Json;
 using System.Threading.Tasks;
-
+using Altinn.Platform.Events.Exceptions;
 using Altinn.Platform.Events.Models;
 using Altinn.Platform.Events.Repository;
 using Altinn.Platform.Events.Services.Interfaces;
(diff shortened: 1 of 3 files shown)


src/Events/Services/AppEventsService.cs
```diff
@@ -66,9 +66,17 @@ public async Task<string> StoreCloudEvent(CloudEvent cloudEvent)
         /// <inheritdoc/>
         public async Task<List<CloudEvent>> GetAppEvents(string after, DateTime? from, DateTime? to, int partyId, List<string> source, List<string> type, string unit, string person, int size = 50)
         {
-            if (partyId <= 0)
+            if ((!string.IsNullOrEmpty(person) || !string.IsNullOrEmpty(unit)) && partyId <= 0)
             {
-                partyId = await _registerService.PartyLookup(unit, person);
+                try
+                {
```
acn-sbuad, line 72: still not sure about this try/catch.. Given that registerService throws the same exception, wouldn't this bubble up to the controller in the first place and not continue the flow further?

SandGrainOne: With the current code I would consider the try-catch to be unnecessary. It does nothing and has no effect.

vgr (approved):
LGTM (test coverage missing approved)

PR #151, opened 2022-10-03
Transfer validated value into subscription object

SandGrainOne:
## Description
When adding the property Validated to Subscription I forgot to ensure the property were set when reading from the result set after DB call.

## Related Issue(s)
- #129


acn-sbuad (approved):
🎈

Issue #28, opened 2022-07-13
Cannot write DateTime with Kind=Unspecified to PostgreSQL type 'timestamp with time zone'


SandGrainOne:
### Description of the bug

Found the following error in Application Insights for Platform in TT02:
`Cannot write DateTime with Kind=Unspecified to PostgreSQL type 'timestamp with time zone', only UTC is supported. Note that it's not possible to mix DateTimes with different Kinds in an array/range. See the Npgsql.EnableLegacyTimestampBehavior AppContext switch to enable legacy behavior. `

Endpoint: `GET /events/api/v1/app/pat/ps-events-test`

[Full call stack.](https://portal.azure.com/#view/AppInsightsExtension/DetailsV2Blade/ComponentId~/%7B%22SubscriptionId%22%3A%22dd6d3e08-a70f-4f71-8847-781ddc5d8468%22%2C%22ResourceGroup%22%3A%22altinnplatform-tt02-rg%22%2C%22Name%22%3A%22tt02-platform-ai%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252Fdd6d3e08-a70f-4f71-8847-781ddc5d8468%252FresourceGroups%252Faltinnplatform-tt02-rg%252Fproviders%252FMicrosoft.Insights%252Fcomponents%252Ftt02-platform-ai%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel~/%7B%22eventId%22%3A%22c374c363-0272-11ed-97e4-29b42f6e9bf1%22%2C%22timestamp%22%3A%222022-07-13T06%3A12%3A03.829Z%22%2C%22cacheId%22%3A%22725ad201-66f3-4479-b8aa-c56f9f2da1ca%22%2C%22eventTable%22%3A%22requests%22%7D)

### Steps To Reproduce

Request parameter example:
https://altinn-tt02-api.azure-api.net/events/api/v1/app/pat/ps-events-test?from=2022-07-07T11:00:53.3917

## Analysis

Assumed to be caused by the from parameter not having defined time zone. In this case the logic should return 400 Bad Request. Return ProblemDetails object with a description of the error.

## Acceptance Criteria

- [ ] Requests with a timestamp value without timezone should give 400. ProblemDetails with error describing the error.

vgr:
TEST: OK!

[...]

PR #158, opened 2022-10-04
Check for dateTime.Kind specified in query params

acn-sbuad:
## Description
- Validating DateTime-objects in controller to avoid exception being thrown by database.
- Moved validation of common query parameters to a private method that throws an exception. Could be middleware instead or set up as a Guard I guess. Open for suggestions here to avoid duplication of the validation logic. 

## Related Issue(s)
- #28

## Verification
~~- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)~~
- [x] All tests run green

### DIFF
--- a/src/Events/Controllers/AppEventsController.cs
+++ b/src/Events/Controllers/AppEventsController.cs
@@ -215,7 +215,7 @@ namespace Altinn.Platform.Events.Controllers
             }
         }
 
-        private void ValidateQueryParams(string after, DateTime? from, DateTime? to, int size)
+        private static void ValidateQueryParams(string after, DateTime? from, DateTime? to, int size)
         {
             if (string.IsNullOrEmpty(after) && from == null)
             {
--- a/src/Events/Controllers/AppEventsController.cs
+++ b/src/Events/Controllers/AppEventsController.cs
@@ -225,7 +225,7 @@ namespace Altinn.Platform.Events.Controllers
 
             if (to != null && to.Value.Kind == DateTimeKind.Unspecified)
             {
-                return (false, "To must specify timezone. E.g. 2022-07-07T11:00:53.3917Z for UTC");
+                return (false, "The 'To' parameter must specify timezone. E.g. 2022-07-07T11:00:53.3917Z for UTC");
             }
 
             if (size < 1)
--- a/src/Events/Controllers/AppEventsController.cs
+++ b/src/Events/Controllers/AppEventsController.cs
@@ -215,12 +215,12 @@ namespace Altinn.Platform.Events.Controllers
         {
             if (string.IsNullOrEmpty(after) && from == null)
             {
-                return (false, "From or after must be defined.");
+                return (false, "The 'From' or 'After' parameter must be defined.");
             }
 
             if (from != null && from.Value.Kind == DateTimeKind.Unspecified)
             {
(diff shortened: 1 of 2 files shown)


src/Events/Controllers/AppEventsController.cs
```diff
@@ -217,6 +215,29 @@ public async Task<ActionResult<string>> Post([FromBody] CloudEventRequestModel c
             }
         }
 
+        private static void ValidateQueryParams(string after, DateTime? from, DateTime? to, int size)
+        {
+            if (string.IsNullOrEmpty(after) && from == null)
+            {
+                throw new ArgumentException("From or after must be defined.");
```
SandGrainOne, line 222: Could we handle this with a return value instead? null if nothing is wrong, a string with an error message if validation fails.

Another, bigger, alternative would be to use ModelState. Collect the parameters in a model and have validation of the model being performed by the model binder. Then just check ModelState.IsValid.

src/Events/Controllers/AppEventsController.cs
```diff
@@ -217,6 +215,29 @@ public async Task<ActionResult<string>> Post([FromBody] CloudEventRequestModel c
...
+            }
+
+            if (from != null && from.Value.Kind == DateTimeKind.Unspecified)
```
vgr, line 225: Do we support non-UTC datetimekinds?

SandGrainOne: Yes, but if we want to prevent a variation of timezones being used stored in the database we might need to actively change the zone to UTC. I haven't thought of it before.

src/Events/Controllers/AppEventsController.cs
acn-sbuad: Tested ModelState, but not happy with the return object. Kept the code in a seperate branch so it can be revisited at a later point. Introducing a touple with validationstatus + error message for now.

src/Events/Controllers/AppEventsController.cs
acn-sbuad: What is specified in Swagger for now is that the time should be defined in UTC, but I guess we can change this is we want. Converting it to UTC before passing it on to the service is an easy task. 
Before we have defined that Platform only uses UTC, and that conversion to local time should be done by the clients / app when required

src/Events/Controllers/AppEventsController.cs
```diff
@@ -217,6 +211,31 @@ public async Task<ActionResult<string>> Post([FromBody] CloudEventRequestModel c
...
+            {
+                return (false, "From must specify timezone. E.g. 2022-07-07T11:00:53.3917Z for UTC");
+            }
+
+            if (to != null && to.Value.Kind == DateTimeKind.Unspecified)
+            {
+                return (false, "To must specify timezone. E.g. 2022-07-07T11:00:53.3917Z for UTC");
```
SandGrainOne, line 228:
```suggestion
                return (false, "The 'To' parameter must specify timezone. E.g. 2022-07-07T11:00:53.3917Z for UTC");
```
Same suggestion for the other messages. It took me a few seconds before I understood the sentence the way it was.

SandGrainOne (approved):
Just one suggestion regarding the messages.

Issue #124, opened 2022-09-19
New controller endpoint and service for storing of event to database


SandGrainOne:
### Description

Storing events and starting pushing of events should be separated operations. As the code is written now, if storing an event in the db fails and throws an exception, the event is lost. This is partly because of the client implementation in apps, but we would like our endpoint to handle this better.

The new Endpoint should support calls from an Azure Function which takes over the responsibility of handling both storage and push of events. This storage endpoint is created because we don't want the Azure Function direct access to the database.

### Consideration
Can we push an event out before it is stored in the database? 
No, events should be stored before being pushed. All pushed events must be available through the GET endpoint.

Maybe we rather should have retry for the events storage to db? 
Yes, there should be rety of both DB calls and queuing.

Both points above is handled by the new design. No extra tasks related to these considerations.

### Tasks

- [x] EventsService is expanded with two new methods for saving and pushing to inbound queue, separately.
- [x] New EventsStorageController for storing of a new event. POST {baseUrl}/storage/events
- [x] New PushToInboundController: POST {baseUrl}/push/inbound 
- [x] Update API Management
- [x] Review/QA

### Acceptance criteria

- [x] Storing an event no longer automatically push an event.
- [x] Pushing an event can be initiated independently of event storage.
- [x] Endpoint access control is verified

vgr:
Manually tested in AT23:
`GET https://altinn-testtools-token-generator.azurewebsites.net/api/GetEnterpriseToken?env=at23&scopes=altinn:serviceowner/instances.read%20altinn:serviceowner/instances.write%20altinn:enduser&org=ttd&orgNo=991825827
GET https://platform.at23.altinn.cloud/events/api/v1/app/ttd/apps-test?from=2014-11-15T18:32:17Z
GET https://platform.at23.altinn.cloud/events/api/v1/app/ttd/apps-test?from=2022-10-18T00:00:01.000000Z

https://ttd.apps.at23.altinn.cloud/ttd/apps-test/#/instance/50334484/6d5bccdc-8614-495b-b537-dfd98b7b3758

POST https://platform.at23.altinn.cloud/events/api/v1/subscriptions
GET https://platform.at23.altinn.cloud/events/api/v1/app/party?from=2022-02-14%2007:22:19Z&party=50334484
GET https://platform.at23.altinn.cloud/events/api/v1/app/ttd/apps-test?from=2022-10-18T00:00:01.000000Z
[...]

PR #161, opened 2022-10-06
Feature/124 save cloudevent to db

bdb-dd:
Added new controllers for saving and pushing events as independent operations. 
First known client will be an Azure Function that processes incoming Cloud Events on the events-registration queue. 
Note: these new endpoints are not yet in use so this pull request is not intended to have any effect on operations. These new endpoints are NOT intended for general use. 

## Related Issue(s)
- #124 

## Verification
- [x] Relevant automated test added
- [x] All tests run green

## Documentation
No documentation changes required for this PR specifically.

### DIFF
--- a/test/Altinn.Platform.Events.Tests/TestingControllers/InboundControllerTests.cs
+++ b/test/Altinn.Platform.Events.Tests/TestingControllers/InboundControllerTests.cs
@@ -34,17 +34,17 @@ namespace Altinn.Platform.Events.Tests.TestingControllers
         /// <summary>
         /// Represents a collection of integration tests of the <see cref="InboundController"/>.
         /// </summary>
-        public class PushControllerInboundTests : IClassFixture<WebApplicationFactory<InboundController>>
+        public class InboundControllerTests : IClassFixture<WebApplicationFactory<InboundController>>
         {
             private const string BasePath = "/events/api/v1";
 
             private readonly WebApplicationFactory<InboundController> _factory;
 
             /// <summary>
-            /// Initializes a new instance of the <see cref="PushControllerInboundTests"/> class with the given <see cref="WebApplicationFactory{TPushController}"/>.
+            /// Initializes a new instance of the <see cref="InboundControllerTests"/> class with the given <see cref="WebApplicationFactory{TPushController}"/>.
             /// </summary>
             /// <param name="factory">The <see cref="WebApplicationFactory{TPushController}"/> to use when setting up the test server.</param>
-            public PushControllerInboundTests(WebApplicationFactory<InboundController> factory)
+            public InboundControllerTests(WebApplicationFactory<InboundController> factory)
             {
                 _factory = factory;
             }


src/Events/Clients/QueueClient.cs
```diff
@@ -2,46 +2,42 @@
...
 
-namespace Altinn.Platform.Events.Services
+namespace Altinn.Platform.Events.Clients
 {
     /// <summary>
     /// The queue service that handles actions related to the queue storage.
     /// </summary>
     [ExcludeFromCodeCoverage]
-    public class QueueService : IQueueService
+    public class QueueClient : IQueueClient
```
SandGrainOne, line 16: I suggest changing the names of the class and interface to EventsQueueClient.

src/Events/Clients/EventsQueueClient.cs
```diff
@@ -2,46 +2,42 @@
...
     /// The queue service that handles actions related to the queue storage.
     /// </summary>
     [ExcludeFromCodeCoverage]
-    public class QueueService : IQueueService
+    public class EventsQueueClient : IEventsQueueClient
     {
         private readonly QueueStorageSettings _settings;
 
-        private QueueClient _inboundQueueClient;
+        private Azure.Storage.Queues.QueueClient _inboundQueueClient;
```
acn-sbuad, line 20: With the name change of the class, I don't think the full name of the queueClient required.

src/Events/Services/Interfaces/IAppEventsService.cs
```diff
@@ -12,11 +12,26 @@ namespace Altinn.Platform.Events.Services.Interfaces
...
         /// <returns>Id for the created document</returns>
-        Task<string> StoreCloudEvent(CloudEvent cloudEvent);
+        Task<string> SaveToDatabase(CloudEvent cloudEvent);
+
+        /// <summary>
+        /// Push cloud event to inbound queue. Should only be attempted after storing event in db.
+        /// </summary>
+        /// <param name="cloudEvent">The cloudEvent to be queued</param>
+        /// <returns>Id for the queued event</returns>
+        Task<string> PushToInboundQueue(CloudEvent cloudEvent);
```
acn-sbuad, line 26: Wondering if these names say too much about the inner workings. Would it be enough for the interface to specify `RegisterEvent`, `StoreEvent` / 'PersistEvent'. The fact that we're using a queue is a part of the implementation but not defining of the interface

src/Events/Clients/Interfaces/IEventsQueueClient.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Pushes the provided content to the queue
+        /// </summary>
+        /// <param name="content">The content to push to the queue in string format</param>
+        /// <returns>Returns a queue receipt</returns>
+        public Task<PushQueueReceipt> PushToOutboundQueue(string content);
+
+        /// <summary>
+        /// Pushes the provided content to the queue
```
acn-sbuad, line 26:
```suggestion
        /// Pushes the provided content to the validation queue
```

src/Events/Controllers/AppEventsController.cs
```diff
@@ -83,7 +83,8 @@ public async Task<ActionResult<string>> Post([FromBody] CloudEventRequestModel c
 
             try
             {
-                string cloudEventId = await _eventsService.StoreCloudEvent(_mapper.Map<CloudEvent>(cloudEvent));
+                string cloudEventId = await _eventsService.SaveAndPushToInboundQueue(_mapper.Map<CloudEvent>(cloudEvent));
+                _logger.LogInformation("Cloud Event successfully stored with id: {cloudEventId}", cloudEventId);
```
acn-sbuad, line 87: Do we need these log messages or we're they just used for development? I guess an exception will be thrown if it isn't successful.

src/Events/Controllers/PushController.cs
```diff
@@ -1,50 +0,0 @@
...
-        /// This method will then identify any matching subscriptions and authorize if the consumer is authorized
-        /// to receive event. If autorized it will put it on a outbound queue
-        /// </remarks>
-        /// <returns>Returns the result of the request in the form og a HTTP status code.</returns>
-        [Authorize(Policy = "PlatformAccess")]
-        [HttpPost]
-        [Consumes("application/json")]
-        [ProducesResponseType(StatusCodes.Status200OK)]
-        [Produces("application/json")]
-        public async Task<ActionResult> Post([FromBody] CloudEvent cloudEvent)
```
acn-sbuad, line 43: is this endpoint moved else where? 
It is currently in use by existing functions, so endpoint shouldn't be deleted before the functions have another available endpoint to use

src/Events/Clients/EventsQueueClient.cs
```diff
@@ -2,22 +2,19 @@
...
-
 using Azure.Storage.Queues;
-
 using Microsoft.Extensions.Options;
 
-namespace Altinn.Platform.Events.Services
+namespace Altinn.Platform.Events.Clients
 {
     /// <summary>
     /// The queue service that handles actions related to the queue storage.
```
SandGrainOne, line 14: The summary still use the term service. Be as specific as you dare to be. Something like:
```suggestion
    /// Implementation of the <see ref="IEventsQueueClient"/> using Azure Storage Queues.
```

src/Events/Controllers/PushToInboundController.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Altinn.Platform.Events.Controllers
+{
+    /// <summary>
+    /// Provides operations for handling events
```
SandGrainOne, line 20: `handling events` could be more specific.

src/Events/Controllers/PushToInboundController.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    [Authorize]
+    [Route("events/api/v1/push/inbound")]
```
SandGrainOne, line 23: I'm not sure it's safe to rename the endpoint like this. What's the plan?

SandGrainOne (commented):
Some initial comments

src/Events/Controllers/PushToInboundController.cs
acn-sbuad: inboud should be a part of the route directly on the method. 
So  [Route("events/api/v1/push")] on the controller

And on the method     [Route("inbound")]

src/Events/Clients/Interfaces/IEventsQueueClient.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        /// <summary>
+        /// Pushes the provided content to the queue
+        /// </summary>
+        /// <param name="content">The content to push to the queue in string format</param>
+        /// <returns>Returns a queue receipt</returns>
+        public Task<PushQueueReceipt> PushToInboundQueue(string content);
+
+        /// <summary>
+        /// Pushes the provided content to the queue
```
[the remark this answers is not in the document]
acn-sbuad: Cant see my suggestion here for some reason. It was to change the summarty to Push the ... to the outbound queue

src/Events/Controllers/AppEventsController.cs
acn-sbuad: I would think the same :) I would suggest we delete the logging of info

src/Events/Controllers/PushToInboundController.cs
SandGrainOne: I'm worried about the removal of `/events/api/v1/push`. It's being replaced with `events/api/v1/push/inbound`, but I want a plan for that change.

SandGrainOne: ohh!! Okay, the new endpoint is supposed to be in addition to the one we've removed.

src/Events/Controllers/EventsStorageController.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <param name="cloudEvent">The cloudEvent to be saved</param>
+        /// <returns>The application metadata object.</returns>
+        [Authorize(Policy = "PlatformAccess")]
+        [HttpPost]
+        [Consumes("application/json")]
+        [SwaggerResponse(201, Type = typeof(Guid))]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [Produces("application/json")]
+        public async Task<ActionResult<string>> Post([FromBody] CloudEventRequestModel cloudEvent)
```
acn-sbuad, line 51: Shouldn't this be a cloudEvent rather than RequestModel? CloudEventRequestModel is used for the POST /app and POST /events endpoints

acn-sbuad: Or even a different model e.g. CloudRegistrationModel/CloudRegistrationEnvelope that also contains a property indicating whether storing the event to the database was successful.

src/Events/Controllers/EventsStorageController.cs
```diff
@@ -0,0 +1,71 @@
...
+using Altinn.Platform.Events.Services.Interfaces;
+using AutoMapper;
+
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Altinn.Platform.Events.Controllers
```
acn-sbuad, line 13: In the series of naming things. Could the file and controller name match the endpoint? -> StorageController

src/Events/Controllers/PushController.cs
```diff
@@ -1,50 +0,0 @@
-using System.Threading.Tasks;
-
-using Altinn.Platform.Events.Models;
-using Altinn.Platform.Events.Services.Interfaces;
-
-using Microsoft.AspNetCore.Authorization;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Mvc;
-
-namespace Altinn.Platform.Events.Controllers
```
acn-sbuad, line 10: This file should be reinstated

src/Events/Controllers/EventQueueController.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        /// <returns>The application metadata object.</returns>
+        [Authorize(Policy = "PlatformAccess")]
+        [HttpPost("inbound")]
+        [Consumes("application/json")]
+        [SwaggerResponse(201, Type = typeof(Guid))]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [Produces("application/json")]
+        public async Task<ActionResult<string>> Post([FromBody] CloudEventRequestModel cloudEvent)
```
acn-sbuad, line 72: Input model here should not be CloudEventRequestModel, ref comment in StorageController

src/Events/Services/Interfaces/IAppEventsService.cs
```diff
@@ -12,14 +12,30 @@ namespace Altinn.Platform.Events.Services.Interfaces
...
-        Task<string> StoreCloudEvent(CloudEvent cloudEvent);
+        Task<string> SaveToDatabase(CloudEvent cloudEvent);
+
+        /// <summary>
+        /// Push cloud event to inbound queue.
+        /// Should only be attempted after saving to persistent storage. 
+        /// </summary>
+        /// <param name="cloudEvent">The cloudEvent to be queued</param>
+        /// <returns>Id for the queued event</returns>
+        Task<string> PushToInboundQueue(CloudEvent cloudEvent);
```
acn-sbuad, line 27: Pushing to inbound queue is agnostic w.r.t. event type, so maybe this would be the time to create a new IEventService which handles push to InboundQueue that can be reused for external events as well ?

src/Events.Functions/Clients/OutboundClient.cs
```diff
@@ -6,54 +6,54 @@
...
 using Microsoft.Extensions.Options;
 
-namespace Altinn.Platform.Events.Functions.Services
+namespace Altinn.Platform.Events.Functions.Clients
 {
     /// <summary>
     /// Handles PushEvents service
     /// </summary>
-    public class PushEventsService : IPushEventsService
+    public class OutboundClient : IOutboundClient
```
acn-sbuad, line 22: I would suggest limiting this PR to changes in the events app, and leaving the function changes for the issue related to the Azure Function.

src/Events/Services/InboundService.cs
```diff
@@ -48,17 +48,45 @@ public class AppEventsService : IAppEventsService
         }
 
         /// <inheritdoc/>
-        public async Task<string> StoreCloudEvent(CloudEvent cloudEvent)
+        public async Task<string> Save(CloudEvent cloudEvent)
```
vgr, line 51: Is the need for SaveAndPostInbound only to support the old app-event implementation, or will we use it going forward? Should we document the difference in functionality (that SaveAndPost creates the id) in the function documenation?

src/Events/Controllers/InboundController.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Post a previously registered cloudEvent to the inbound events queue
+        /// and further processing by the EventsInbound Azure function
+        /// </summary>
+        /// <returns>The application metadata object.</returns>
+        [Authorize(Policy = "PlatformAccess")]
+        [HttpPost]
+        [Consumes("application/json")]
+        [SwaggerResponse(201, Type = typeof(Guid))]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
```
vgr, line 39: How does this return 400? It seems that if PostInbound fails to enqueue an event this will still return 201 Created?

src/Events/Clients/Interfaces/IEventsQueueClient.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Altinn.Platform.Events.Clients.Interfaces
+{
+    /// <summary>
+    /// Describes the necessary methods for an implementation of an events queue client.
+    /// </summary>
+    public interface IEventsQueueClient
+    {
+        /// <summary>
+        /// Pushes the provided content to the inbound queue
```
acn-sbuad, line 12:
```suggestion
        /// Enqueues the provided content to the inbound queue
```

src/Events/Clients/Interfaces/IEventsQueueClient.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// </summary>
+        /// <param name="content">The content to push to the queue in string format</param>
+        /// <returns>Returns a queue receipt</returns>
+        public Task<QueuePostReceipt> EnqueueInbound(string content);
+
+        /// <summary>
+        /// Pushes the provided content to the outbound queue
```
acn-sbuad, line 19:
```suggestion
        /// Enqueues the provided content to the outbound queue
```

src/Events/Services/Interfaces/IInboundService.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// pushes to Inbound events queue in one operation. 
+        /// </summary>
+        /// <param name="cloudEvent">The cloudEvent to be stored</param>
+        /// <returns>Id for the cloudEvent in persistent storage</returns>
+        Task<string> SaveAndPostInbound(CloudEvent cloudEvent);
+
+        /// <summary>
+        /// Gets list of cloud events based on query params
+        /// </summary>
+        Task<List<CloudEvent>> GetAppEvents(string after, DateTime? from, DateTime? to, int partyId, List<string> source, List<string> type, string unit, string person, int size = 50);
```
acn-sbuad, line 40: Not sure I like the moving of this function. Both this and the save function might belong in the deleted appEvents service interface for now. Only methods related to the inbound processing of instances should be defined in this interface ace.

src/Events/Controllers/StorageController.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <summary>
+    /// Provides operations for saving and retrieving cloud events from persistent storage.
+    /// </summary>
+    [Authorize]
+    [Route("events/api/v1/storage/events")]
+    public class StorageController : ControllerBase
+    {
+        private readonly IInboundService _inboundService;
+        private readonly ILogger _logger;
+        private readonly IMapper _mapper;
```
acn-sbuad, line 24: Not sure if you need a mapper here? This controller should only accept and process objects of the cloudEvent model

src/Events/Controllers/PushController.cs
```diff
@@ -1,50 +0,0 @@
...
-{
-    /// <summary>
-    /// Controller responsible for pushing events to subscribers
-    /// </summary>
-    [Route("events/api/v1/push")]
-    [ApiController]
-    public class PushController : ControllerBase
```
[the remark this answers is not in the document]
acn-sbuad: The push endpoint is still in use by the function and should not be removed untill we have another endpoint replacing it.

src/Events/Controllers/OutboundController.cs
```diff
@@ -0,0 +1,78 @@
...
+            return Ok();
+        }
+
+        /// <summary>
+        /// Push a previously registered cloudEvent to the inbound events queue
+        /// and further processing by the EventsInbound Azure function
+        /// </summary>
+        /// <returns>The application metadata object.</returns>
+        [Authorize(Policy = "PlatformAccess")]
+        [HttpPost("inbound")]
```
acn-sbuad: obs, inbound here might be a copy/paste error ?

acn-sbuad: or, looking closer, this might just be duplicated from the inbound controller and can be deleted :)

src/Events/Services/Interfaces/IInboundService.cs
vgr: I don't really think of it as an inbox and outbox. That is in my mind a concept related to messages, which in addition to data have a sender and a receiver.

> An [event](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#event) includes context and data about an [occurrence](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#occurrence). Each occurrence is uniquely identified by the data of the event.
>Events represent facts and therefore do not include a destination, whereas messages convey intent, transporting data from a source to a given destination.
>https://github.com/cloudevents/spec/blob/main/cloudevents/primer.md

SandGrainOne: With all these queues and operations to handle them we're breaking up the traditional layered application we're used to working with. One trick could be to do a virtual microservice split. Which applications would we get and what would each of them be responsible for? What would we change if we decided to make events into two applications?

I personally feel like the logic of handling everything related to sending events to subscribers could be a separate application. I my head this respinsibility starts when something appears in the events-inbound queue, or when the application is made aware of a new event. 

The other stuff related to new incomming events, having them stored, retrieved, listed/filtered could be an other application.

vgr: I think of inbound as the start of the subscription capabilities of our system. Hypothetically, if we had zero subscribers it could make sense to pause/stop the inbound queue, since there is no need to send information about events to subscribers.

So while sending an event to subscibers definitively can be viewed as our event system's "outbox", I don't think that creating an event is the "inbox". That is just registering an event, in my head. The event don't become a message with a destination before we have subscribers, and that is something that happens after registering and creating an event.

src/Events/Services/Interfaces/IOutboundService.cs
```diff
@@ -7,11 +7,11 @@ namespace Altinn.Platform.Events.Services.Interfaces
     /// <summary>
     /// Defines the methods required for an implementation of the push service.
     /// </summary>
-    public interface IPushEvent
+    public interface IOutboundService
     {
         /// <summary>
-        /// Push a event to all consumer subcribing to source and/or subject
+        /// Deliver inbound cloudEvent to each matching subscriber.
```
acn-sbuad, line 13:
```suggestion
        /// Deliver outbound cloudEvent to each matching subscriber.
```

src/Events.Functions/Clients/Interfaces/IOutboundClient.cs
```diff
@@ -0,0 +1,17 @@
+using System.Threading.Tasks;
+using Altinn.Platform.Events.Functions.Models;
+
+namespace Altinn.Platform.Events.Functions.Clients.Interfaces
+{
+    /// <summary>
+    /// Interface to handle services exposed in Platform Events Push
```
acn-sbuad, line 7: Summary here should be updated

src/Events.Functions/Clients/OutboundClient.cs
```diff
@@ -6,54 +6,54 @@
...
 
         /// <inheritdoc/>
-        public async Task SendToPushController(CloudEvent item)
+        public async Task PostOutbound(CloudEvent item)
         {
             StringContent httpContent = new(JsonSerializer.Serialize(item), Encoding.UTF8, "application/json");
             try
             {
-                string endpointUrl = "push";
+                string endpointUrl = "outbound";
```
acn-sbuad: You've changed the output here.. Meaning we risk that the endpoint isn't available when the functions is deployed. It is my strong recommendation that all changes related to Events.Functions is removed from this PR and handled separately.

src/Events/Services/Interfaces/IInboundService.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+using Altinn.Platform.Events.Models;
+
+namespace Altinn.Platform.Events.Services.Interfaces
+{
+    /// <summary>
+    /// Interface to talk to the events service
```
acn-sbuad, line 10: Could we get a more precise summary?

src/Events/Services/Interfaces/IInboundService.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Push cloud event to inbound queue.
+        /// Should only be attempted after saving to persistent storage. 
+        /// </summary>
+        /// <param name="cloudEvent">The cloudEvent to be queued</param>
+        /// <returns>Id for the queued event</returns>
+        Task<string> PostInbound(CloudEvent cloudEvent);
+
+        /// <summary>
+        /// [DEPRECATED] Save a cloud event to persistent storage and
```
acn-sbuad, line 30: To ensure that the information about the deprecation showing up in build logs etc. you should use `[Obsolete("message")]`
`

src/Events/Controllers/OutboundController.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Produces("application/json")]
+        public async Task<ActionResult> Post([FromBody] CloudEvent cloudEvent)
+        {
+            try
+            {
+                await _outboundService.PostOutbound(cloudEvent);
+
+                return Ok();
+            }
+            catch (Exception e)
```
acn-sbuad, line 52: Catching the exception here, but not logging it means we miss out on information of what caused the exception in our logs.

src/Events/Controllers/InboundController.cs
```diff
@@ -0,0 +1,55 @@
...
+        public InboundController(IInboundService inboundService)
+        {
+            _inboundService = inboundService;
+        }
+
+        /// <summary>
+        /// Post a previously registered cloudEvent to the inbound events queue
+        /// and further processing by the EventsInbound Azure function
+        /// </summary>
+        /// <returns>The application metadata object.</returns>
```
acn-sbuad, line 34: Returns here must be updated

src/Events/Controllers/InboundController.cs
```diff
@@ -0,0 +1,55 @@
...
+        public async Task<ActionResult<string>> Post([FromBody] CloudEvent cloudEvent)
+        {
+            try
+            {
+                string cloudEventId = await _inboundService.PostInbound(cloudEvent);
+
+                return Created(cloudEvent.Subject, cloudEventId);
+            }
+            catch (Exception e)
+            {
```
acn-sbuad, line 50: Same goes for loss of exception information in our logs

src/Events/Controllers/OutboundController.cs
acn-sbuad: No, if it's logget in the service that should be enoug.

src/Events/Controllers/InboundController.cs
SandGrainOne: There is a lot of code in this PR that appear to be new, but isn't. This is one example. I've allowed @b-brodie-digdir to focus on finishing up the actual focus of the PR and ignore old mistakes this time around.

SandGrainOne (approved):

vgr (approved):

Issue #112, opened 2022-09-09
Create function that reads new events from queue and, store and push the events


SandGrainOne:
### Description

We want to make the act of storing and starting push of an event asynchornous. This will be done by a new Azure Function.

pseudo-code:
1. Trigger based on new events on the "storage" queue: Name: events-registration
2. Check whether the events has been stored already, store if not
3. Put the event on to the push queue (EventsInbound). 
    1. If this fails. Put the event back on the "event-registration" queue with a status indicating that the event is stored

![storage queue drawio](https://user-images.githubusercontent.com/2217340/193529237-e80a364f-daa3-4874-8969-f6177ea36033.svg)

### Additional Information

The function will need an AccessToken / PlatformAccess to call endpoints in the events API.

### Tasks

- [x] Create new Azure Function. EventsRegistration.
~~- [ ] Create a new test project for unit tests of functions.~~ Postponing development of test projects to later issue as complexity of implementation decreased during development.
~~- [ ] Update github workflows as necessary to run the tests.~~

~~- [ ] Automatic testing? UseCase tests?~~ handled elsewhere
- [x] Review/QA
~~- [ ] Documentation (docs/pedia)~~ documentation handled during #118

### Acceptance Criterias

- [x] Events on the queue is being stored and pushed

PR #162, opened 2022-10-06
Feature/112 process events registration queue

b-brodie-digdir:
## EventsRegistration
 - new Azure function with QueueTrigger based activation.

### Primary responsibility: 
- save incoming cloud events to persistent storage (db)

### Process:
1. dequeue Cloud Event from `events-registration` queue
2. save to internal persistent storage (postgres) 
    - POST {baseUrl}/storage/events 
3. enqueue events-inbound queue via 
   - POST {baseUrl}/inbound

### Related Issue(s)
- #112 
- #124 

## Verification
- [x] Code builds clean without any errors or warnings
- [ ] All tests run green

## Documentation
- [ ] Update process diagram to reflect new `events-registration` queue and this function.


src/Events.Functions/Clients/EventsStorageClient.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        private readonly HttpClient _client;
+        private readonly IAccessTokenGenerator _accessTokenGenerator;
+        private readonly IKeyVaultService _keyVaultService;
+        private readonly KeyVaultSettings _keyVaultSettings;
+        private readonly PlatformSettings _platformSettings;
+        private readonly ILogger<IPushToOutboundQueueClient> _logger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PushToOutboundQueueClient"/> class.
```
acn-sbuad, line 32:
```suggestion
        /// Initializes a new instance of the <see cref="EventsStorageClient"/> class.
```

src/Events.Functions/Clients/EventsStorageClient.cs
```diff
@@ -0,0 +1,80 @@
...
+                    "platform",
+                    "events",
+                    new X509Certificate2(Convert.FromBase64String(certBase64), (string)null, X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable));
+
+                HttpResponseMessage response = await _client.PostAsync(endpointUrl, httpContent, accessToken);
+                if (response.StatusCode != HttpStatusCode.OK)
+                {
+                    var msg = $"// SaveCloudEvent with id {item.Id} failed with statuscode {response.StatusCode}";
+                    _logger.LogError(msg);
+                    throw new Exception(msg);
```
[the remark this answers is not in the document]
acn-sbuad: What's we've done for other clients communicating with platform is that we've created a custom exception type PlatformHttpException (I think) and then relayed the status code from the platform component. That could be done here as well. If not the HttpRequestException seems like the better option

src/Events.Functions/Clients/EventsStorageClient.cs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, $"// SaveCloudEvent with id {item.Id} failed with errormessage {e.Message}");
+                throw e;
```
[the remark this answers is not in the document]
acn-sbuad: Not done to hide stack information. Would be to allow the function to fail after the exception has been logged on our part. 
```suggestion
                throw;
```

src/Events/Services/AppEventsService.cs
```diff
@@ -48,17 +48,45 @@ public class AppEventsService : IAppEventsService
         }
 
         /// <inheritdoc/>
-        public async Task<string> StoreCloudEvent(CloudEvent cloudEvent)
+        public async Task<string> SaveToDatabase(CloudEvent cloudEvent)
+        {
+            try
+            {
+                cloudEvent = await _repository.Create(cloudEvent);
```
vgr, line 55: C#/.Net question: Create overwrites cloudEvent. If it fails, ex: casts exception during deserialization of the cloud event coming from db, will the variable be overwritten by something that makes cloudEvent.Id potentially fail? Or is this safe behaviour when working with async functions in .Net?

src/Events.Functions/Clients/EventsStorageClient.cs
```diff
@@ -0,0 +1,80 @@
...
+using Altinn.Platform.Events.Functions.Services.Interfaces;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Altinn.Platform.Events.Functions.Clients
+{
+    /// <summary>
+    /// Events Storage REST API client
+    /// </summary>
+    public class EventsStorageClient : IEventsStorageClient
```
vgr, line 22: There is a lot of shared code between the new clients. Could we maybe share more of it by creaing an astract class that handles generateAccessToken?

src/Events.Functions/Clients/EventsStorageClient.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        private readonly HttpClient _client;
+        private readonly IAccessTokenGenerator _accessTokenGenerator;
+        private readonly IKeyVaultService _keyVaultService;
+        private readonly KeyVaultSettings _keyVaultSettings;
+        private readonly PlatformSettings _platformSettings;
```
vgr, line 28: PlatformSettings can become a local variable.

src/Events.Functions/Clients/EventsStorageClient.cs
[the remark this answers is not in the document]
acn-sbuad: My suggestion was to create a seperate exceptin class in this project instead of creating a dependency between the project, but this HttpRequestException is good enough for now.

src/Events.Functions/Clients/SecureClientBase.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    protected IAccessTokenGenerator AccessTokenGenerator { get; }
+
+    /// <summary>
+    /// Access to KeyVault
+    /// </summary>
+    protected IKeyVaultService KeyVaultService { get; }
+
+    /// <summary>
+    /// KeyValue settings
```
vgr, line 33:
```suggestion
    /// KeyVault settings
```

src/Events.Functions/EventsRegistration.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            CloudEvent cloudEvent = JsonSerializer.Deserialize<CloudEvent>(item);
+
+            /*
+            Attempt to save cloudEvent. 
+            If saving fails, the cloudEvent will automatically be 
+            returned to events-registration queue for retry handling.
+            */
+
+            await _eventsStorageClient.SaveCloudEvent(cloudEvent);
```
vgr, line 48: Do we need to handle the potential throw from SaveCloudEvent? eventsStorageClient will log something, but do we need to handle this in the function as well?

b-brodie-digdir: I am concerned we already have logging in triplicate here already. In the case of an http error in the target api, our handlers will log two (identical) errors and the QueueTrigger runtime will also log something. 
On second look, I think we could simplify all clients if we isolate logging of AccessToken generation errors in the base class.

src/Events.Functions/Clients/ValidateSubscriptionClient.cs
```diff
@@ -4,43 +4,40 @@
...
 using Microsoft.Extensions.Options;
 
-namespace Altinn.Platform.Events.Functions.Services
+namespace Altinn.Platform.Events.Functions.Clients
 {
     /// <summary>
     /// Service to validate subscription
     /// </summary>
-    public class ValidateSubscriptionService : IValidateSubscriptionService
+    public class ValidateSubscriptionClient : SecureClientBase, IValidateSubscriptionClient
```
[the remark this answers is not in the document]
b-brodie-digdir: Good catch. Fixed.

src/Events.Functions/Clients/SecureClientBase.cs
b-brodie-digdir: Auto-correct :-]

src/Events.Functions/EventsRegistration.cs
acn-sbuad: I don't think we need any extra logging or handling here as the event will end up back on the queue.. I do however question the extra load on the events component if the queue is the error source. The component might be idempotent, but do we want this extra load on the solution if the event has already been persisted successfully ?

src/Events.Functions/Clients/Interfaces/IOutboundClient.cs
```diff
@@ -0,0 +1,17 @@
+using System.Threading.Tasks;
+using Altinn.Platform.Events.Functions.Models;
+
+namespace Altinn.Platform.Events.Functions.Clients.Interfaces
+{
+    /// <summary>
+    /// Interface to handle services exposed in Platform Events Push
```
acn-sbuad, line 7:
```suggestion
    /// Interface to handle services exposed in Platform Events Outbound
```

src/Events.Functions/Clients/Interfaces/IInboundClient.cs
```diff
@@ -0,0 +1,17 @@
+using System.Threading.Tasks;
+using Altinn.Platform.Events.Functions.Models;
```
acn-sbuad, line 2: Please check all summaries, seems to be some copy/paste errors

src/Events.Functions/Clients/Interfaces/IEventsStorageClient.cs
```diff
@@ -0,0 +1,18 @@
+using System.Threading.Tasks;
+using Altinn.Platform.Events.Functions.Models;
+
+namespace Altinn.Platform.Events.Functions.Clients.Interfaces
+{
+    /// <summary>
+    /// Interface to handle services exposed in Platform Events Push
```
acn-sbuad, line 7:
```suggestion
    /// Interface to handle services exposed in Platform Events Storage
```

src/Events.Functions/Clients/InboundClient.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// Client used to send cloudEvents to events-inbound queue.
+    /// </summary>
+    public class InboundClient : SecureClientBase, IInboundClient
+    {
+        private readonly ILogger<IInboundClient> _logger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InboundClient"/> class.
+        /// </summary>
+        public InboundClient(
```
acn-sbuad, line 29: There isn't much logic in either of these clients.. Did you consider only having a single eventsClient that exposes a function for sending a request to each of the relevant endpoints?

src/Events.Functions/EventsRegistration.cs
b-brodie-digdir: I think we concluded that if the target queue is unavailable, then the source queue will also be unavailable and our function will not be triggered. Can you think of any scenario where that wouldn't be the case?

src/Events.Functions/Clients/Interfaces/IEventsClient.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Send cloudEvent for outbound processing.
+        /// </summary>
+        /// <param name="item">CloudEvent to send</param>
+        Task PostOutbound(CloudEvent item);
+
+        /// <summary>
+        /// Validates a subscription
+        /// </summary>
+        /// <returns></returns>
```
acn-sbuad, line 33: empty returns

src/Events.Functions/Clients/EventsClient.cs
```diff
@@ -0,0 +1,161 @@
...
+            var platformSettings = eventsConfig.Value;
+            _client.BaseAddress = new Uri(platformSettings.ApiEventsEndpoint);
+            _logger = logger;
+        }
+
+        /// <summary>
+        /// Generate a fresh access token using the client certificate
+        /// </summary>
+        /// <returns></returns>
+        protected async Task<string> GenerateAccessToken(string issuer, string app)
```
acn-sbuad, line 76: for this function in particular issuer and app are always the same. Could maybe be props of the class instead of passing them in for each request?

acn-sbuad:
I like the current client setup much better. Makes the code easier to follow and less files to jump between. Just a few minor comments

acn-sbuad (approved):
🎈

Issue #95, opened 2022-08-29
Unify source filters for person- and organization-subscription


vgr:
#### Discussion
Currently, fetching the correct person- and organization-subscription is handled slightly differently.

When an event is pushed, a person subscription query is created based on the source URI.
The source is shortened to only include the 2 first path segments of the URL.
Ex: https://ttd.apps.altinn.cloud/ttd/apps-test/instances/123 -> https://ttd.apps.altinn.cloud/ttd/apps-test
This source is then used in PostgreSQL to find subscriptions equal to https://ttd.apps.altinn.cloud/ttd/apps-test

Pros:
- Matching is handled by the DB
  
Cons:
- Only works for Altinn Apps and other source filters that use exactly 2 path segments. will need to be changed to support external apps.

When an event is pushed the organization query does not take into account the source when fetching subscriptions.
Instead this is handled by the code afterwards.

Pros:
- Easily extendable to new URI schemes.
- Source matching is done correctly for URL path segments. (Ref: Issue #25) This is not currently an issue for the other approach, but care needs to be taken when new schemes are supported.

Cons:
- DB result contains all subscriptions. Performance/scalability

#### Expected behaviour:
| event source | event source filter | match? |
| --- | --- | --- |
| https://host.com/a/b/c    | https://host.com/a/b      | yes |
| https://host.com/a/b      | https://host.com/a/b      | yes |
| https://host.com/A/B      | https://host.com/a/b      | ? |
| https://host.com/a/b/c    | https://host.com/a/b/b    | no |
| https://host.com/a/bb     | https://host.com/a/b      | no |
| https://host.com/a/b      | https://host.com/a/bb     | no |
| https://host.com/a/b      | https://host.com/a/b/c    | no |
| https://host.com/a         | https://host.com/a          | not legal  (#96)|

#### Proposed solution
- Use DB matching for all subscriptions
- Consolidate both SQL functions into 1 that retuns both orgs and consumer subscriptions.

The implementation of the SQL function should take into account #96, so we don't need to re-implement it entirely.

### Additional Information
Support for arbitrary URLs have been moved to a new issue #96 

### Tasks

- [x] Ensure that stored data is in lower case
- [x] Ensure that filter values used in where clause  is lower case
- [x] Create unified function to identify subscriptions that should receive an event through push
- [x] Update repository logic as needed
- [x]  Review/QA
- [x] Unit tests
~~- [ ] Documentation~~
(Automated tests moved to separate issue #166 )

[...]

Issue #96, opened 2022-08-31
Add support for external URL and URN source subscriptions


vgr:
### Implementation
See proof-of-concept: https://dotnetfiddle.net/DP2nY8
1. Create a hash of the subscription source filter. -> New column in DB table.
2. When getting all matching subscriptions for an event we can create a set of hashes - one hash for each substring of the event source.
    - "www.host.com/a/b/c/d/e" -> `[ hash("www.host.com"), hash(""www.host.com/a"), hash("www.host.com/a/b", ...]`
    - Can trim `\` from URLs
    - Can use SHA, unless we find a better alternative 
3. SQL function can then use  `WHERE subscription.source IN hash-set`, which allows us to use b-tree index on the hash-column.

### Tasks

- [x] Add new column to subscription table and create an index for this
- [x] Decide algorithm and create hash function
  - MD5 Hash is used by PostgreSQL 
- [x] Generate hash for all existing subscriptions and fill column with values (all environments)
- [x] Update POST subscription endpoint to store hashes when new subscriptions are stored
- [x] Update push to use hashes for subscription-lookup
- [x] Documentation
- [x] Testing

### Acceptance criteria
- [x] Subset matches should not occur. 
- [x] All subscriptions have a new populated colum sourcefilterhash in db 
- [x] Happy path should work as expected: push events
- [x] New subscriptions should have sourcefilterhash populated in db

vgr:
Alternative implementation: https://dotnetfiddle.net/DP2nY8
The hash-set should be more compatible with B-tree indexes since the query can use the IN-operator.

PR #171, opened 2022-10-13
Unify flow for identifying subscription for outbound events 

acn-sbuad:
## Description
- Unified workflows regardless of org or party consumer in subscription.

Remaining: 
- Tweak function to pass all conditions
- Unit test

## Related Issue(s)
- #95

## Verification
~~- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)~~
- [x] All tests run green


src/Events/Migration/v0.14/01-setup-functions.sql
```diff
@@ -0,0 +1,32 @@
...
+    ROWS 1000
+
+AS $BODY$
+
+BEGIN
+return query 
+	SELECT s.id, s.sourcefilter, s.subjectfilter, s.typefilter, s.consumer, s.endpointurl, s.createdby, s.validated, s."time"
+	FROM events.subscription s
+  WHERE (s.subjectfilter is NULL OR s.subjectfilter = subject)
+  AND position(s.sourcefilter in source) = 1
```
vgr, line 27: I was wrong when I said that we should implement `position(subscription.source_filter in event.source) = 1` before we had match-format with brackets.
We should change this (back) to `s.sourcefilter = source` until #96 .

The problem is that without match-format we will incorrectly match
subscription source-filter = `apps.altinn.cloud/ttd/appname`
with an event from source = `apps.altinn.cloud/ttd/appname-print`.

If we reverse the parameter ordering we will match
subscription source-filter = `apps.altinn.cloud/ttd/appname-print`
with an event from source = `apps.altinn.cloud/ttd/appname`.

So, until #96 we should just keep it as an strict `=` since we have build the source explicitly to have 2 path-segments.

src/Events/Services/PushEventService.cs
```diff
@@ -183,34 +149,36 @@ private static CloudEventEnvelope MapToEnvelope(CloudEvent cloudEvent, Subscript
...
         }
 
-        private static string GetAltinnAppAuthorizationCacheKey(string sourceFilter, string consumer)
+        private string GetSourceFilter(Uri source)
         {
-            if (sourceFilter == null)
+            if (source.DnsSafeHost.Contains(_platformSettings.AppsDomain))
             {
-                return null;
+                return source.OriginalString.Substring(0, source.OriginalString.IndexOf(source.Segments[3]) - 1);
```
vgr, line 176: Is there a check that source has at least 3 segments somewhere before this is executed, or are we relying on the creation of app events to only create events with org/app/instance sources?

acn-sbuad: No checks on the segment. As the method is only called from AuthorizeConsumerForAltinnAppEvent, we rely on it being n app event for now

src/Events/Services/OutboundService.cs
```diff
@@ -20,43 +23,38 @@ public class OutboundService : IOutboundService
...
-        private readonly MemoryCacheEntryOptions _partySubscriptioncacheEntryOptions;
+        private readonly MemoryCacheEntryOptions _subscriptionCacheEntryOptions;
         private readonly MemoryCacheEntryOptions _orgAuthorizationEntryOptions;
 
         private readonly ILogger<IOutboundService> _logger;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="OutboundService"/> class.
-        /// </summary>
+        /// </summary>
```
vgr, line 38:
```suggestion
        /// </summary>
```

vgr (approved):

Issue #103, opened 2022-08-26
Utvikle endepunkt for publisering av "eksterne" events


OddArneSaetervik:

SandGrainOne:
Discuss with Authorization the names for the Resource identification parameters.

SandGrainOne:
@TheTechArch Kan du avklare for oss hvordan vi skal identifisere en resurs som ekstern event?

vgr:
<table>
<tr>
	<td>No authorization
	<td> 401 unauthorized
	<td>
<tr>
	<td>Wrong scope
	<td>403 forbidden
	<td>
<tr>
	<td>Minimal event
	<td>201 Created
	<td>containing id, type, source, subject, specversion
<tr>
	<td>"id": null
	<td>400 Bad request
	<td>"CloudEvent is missing required attributes: id (Parameter 'data')"
<tr>
	<td>"source": "a"
	<td>201 created
	<td>
<tr>
	<td>"source": null
	<td>400 Bad request
	<td>"CloudEvent is missing required attributes: source (Parameter 'data')"
<tr>
	<td>
	<td>
	<td>
<tr>
	<td>
	<td>
	<td>
<tr>
	<td>
	<td>
	<td>
<tr>
	<td>
	<td>
	<td>
<tr>
	<td>
	<td>
	<td>
</table>

vgr:
`"subject": null` -> 200 OK
Testet ok!

PR #184, opened 2022-10-24
POST endpoint for external events

acn-sbuad:
## Description

- Using CloudNative CloudEvents SDK
- All exposed cloudEvents follow the spec
- New endpoint for registering events from external sources

- No breaking changes for internal endpoints used by functions in this first release. Will follow up with those.

## Related Issue(s)
- #103 

## Verification
- [x] All tests run green

## Documentation


src/Events/Models/CloudEvent.cs
```diff
@@ -90,5 +90,25 @@ public static CloudEvent Deserialize(string jsonString)
...
         }
+
+        /// <summary>
+        /// Validated required properties Id, Time, Source, SpecVersion, Type and Subject
+        /// </summary>
+        /// <returns>A boolean indicating whether all required fields have a value</returns>
+        public bool ValidateRequiredProperties()
+        {
+            if (string.IsNullOrEmpty(Id) ||
+                Time == null ||
```
b-brodie-digdir, line 101: Looks like `time` should be optional. I'm not sure how we can ensure conformence with RFC 3339 without making use of the cloudevents C# SDK.
https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time

test/Altinn.Platform.Events.Tests/TestingControllers/EventsControllerTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        public class EventsControllerTests : IClassFixture<WebApplicationFactory<EventsController>>
+        {
+            private const string BasePath = "/events/api/v1";
+
+            private readonly WebApplicationFactory<EventsController> _factory;
+            private readonly CloudEventRequestModel _invalidEvent;
+            private readonly CloudEventRequestModel _validEvent;
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="AppControllerTests"/> class with the given <see cref="WebApplicationFactory{TAppEventsController}"/>.
```
b-brodie-digdir, line 48: [AppController|AppEventsController] -> EventsController

test/Altinn.Platform.Events.Tests/TestingServices/EventsServiceTest.cs
```diff
@@ -38,6 +41,10 @@ public EventsServiceTest()
             _authorizationMock = new();
             _claimsPrincipalProviderMock = new();
             _loggerMock = new();
+            _serializerOptions = new()
+            {
+                PropertyNamingPolicy = JsonNamingPolicy.CamelCase
```
b-brodie-digdir, line 46: Found this interesting point in the spec regarding casing:
https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#attribute-naming-convention

Don't have to do it now, but it would be great to look more into what future pain we can avoid by using the cloudevents.io SDK rather than handling serialization ourselves.

SandGrainOne: A small note: The configuration used here will be overridden by the attributes decorating each property in the model(s).

acn-sbuad: using a new jsonSerializer instead. Not possible to set to null in the validation within a mock implementation

src/Events/Models/CloudEventEnvelope.cs
```diff
@@ -13,7 +10,7 @@ public class CloudEventEnvelope
         /// <summary>
         /// The Event to push
         /// </summary>
-        public CloudEvent CloudEvent { get; set; }
+        public object CloudEvent { get; set; }
```
b-brodie-digdir, line 13: I don't recall, did we find a good way to test this change?

src/Events/Program.cs
```diff
@@ -7,14 +7,17 @@
 using Altinn.Common.AccessToken;
```
b-brodie-digdir, line 7: Ref github actions analysis, CogComplexity, should we consider moving some logic to a Startup.cs class?

src/Events/Services/OutboundService.cs
```diff
@@ -147,13 +152,17 @@ private static CloudEventEnvelope MapToEnvelope(CloudEvent cloudEvent, Subscript
...
             CloudEventEnvelope cloudEventEnvelope = new CloudEventEnvelope()
             {
-                CloudEvent = cloudEvent,
                 Consumer = subscription.Consumer,
                 Pushed = DateTime.Now,
                 SubscriptionId = subscription.Id,
                 Endpoint = subscription.EndPoint
             };
 
+            var formatter = new JsonEventFormatter();
```
b-brodie-digdir, line 161: I guess we could use `cloudEvent.SerializeCloudEvent()` here

src/Events/Program.cs
acn-sbuad: We consolidated these two files when migrating to .net 6. Thinking the complexity will be the same if we move it to a Startup.cs. Creating extension methods for registration of services might help with the error messages, but not sure if it is less or more complex to follow the code then. Would suggest we create an issue to analyse, but leave this file as is for now

test/Altinn.Platform.Events.Tests/Data/events/1.json
```diff
@@ -4,35 +4,33 @@
     "id": "e31dbb11-2208-4dda-a549-92a0db8c7708",
     "source": "https://ttd.apps.altinn.no/ttd/endring-av-navn-v2/instances/1337/6fb3f738-6800-4f29-9f3e-1c66862656cd",
     "subject": "/party/1337",
```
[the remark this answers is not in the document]
acn-sbuad: Tested.. We still have tests dependent on this.

b-brodie-digdir (approved):

src/Events/Models/CloudEventEnvelope.cs
acn-sbuad: I tried testing last night, but didn't get very far.  My suggestion would be that we keep cloudEvent as the type for now, and rather update this in the issues related to pushing external events. We don't have support for generating subscriptions, so no cloud envelope will be generated for external events with potential additional params yet.

tba76 (approved):
🍺

Issue #156, opened 2022-10-04
Update event storage endpoint to store generic events to new table


SandGrainOne:
### Description

External and App events should both be stored to a new table. events.events (Requires a rename of existing events.events to events_app. #120 )

### Considerations

- Store the full json and index it in the database or store the search values in individual column separate columns for search values
  - If supported, indexing the json document and querying it would lighten the cognitive load for the developers. 
  - jsonb might be the property type to use
  - App events should be stored to both the new and exsisting table.
  - Generic event should be stored only to the new table.

### Specification
- [x] Define the schema for the events table taking all considerations into account. (#99) 
- [x] New auto populated time column for when the event is stored. Used by the cleanup job.
- [x] Adjust the "schema" to exclude the special time column for the event time.

#### Migration script

```sql
CREATE TABLE IF NOT EXISTS events.events (
    sequenceno BIGSERIAL,
    cloudevent JSONB NOT NULL,
    "registeredtime" timestamptz default (now() at time zone 'utc'),
    CONSTRAINT events_pkey PRIMARY KEY (sequenceno)
);

CREATE INDEX IF NOT EXISTS idx_gin_events_computed_cloudevent ON events.events USING GIN (cloudevent jsonb_path_ops);

CREATE INDEX IF NOT EXISTS idx_events_computed_time ON events.events USING btree ("registeredtime" ASC NULLS LAST);

```

#### Measuring performance
Tried out different type of indexing. Gin indexing jsonb column performed as well as search in the original events-table. Support for search on time must be done with own column since jsonb don't include this datatype. 
Created a computed column for time. This will be populated when inserting json in jsonb-column. 
Since convert from text to timestamp is not immutable, we need an own function to do this.

Sql for testing:
``` sql
create or replace function events.to_timestamptz(timestring text) 
returns timestamptz 
language 'plpgsql'
immutable 
as $BODY$
begin
  return timestring::timestamptz;
end;
$BODY$;

CREATE TABLE IF NOT EXISTS events.events_json_computed_column (
    sequenceno BIGSERIAL,
    cloudevent JSONB NOT NULL,
    "time" timestamptz generated always AS (events.to_timestamptz(cloudevent->>'time')) STORED,
	CONSTRAINT events_json_computed_pkey PRIMARY KEY (sequenceno)
);

CREATE INDEX IF NOT EXISTS idx_gin_events_computed_cloudevent ON events.events_json_computed_column USING GIN (cloudevent jsonb_path_ops);

[...]

PR #208, opened 2022-11-09
Implementation for persisting cloud events in events table

acn-sbuad:
## Description

Service, repository and DB are now ready to support external events. 
Also made changes to ensure app events are stored in both tables. 

## Related Issue(s)
- #156

## Verification
~~- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)~~ will be handled when controller in place
- [x] All tests run green

## Documentation


src/Events/Migration/v0.21/01-setup-table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS events.events (
+    sequenceno BIGSERIAL,
+    cloudevent JSONB NOT NULL,
+    registeredtime timestamptz default (now() at time zone 'utc'),
+    CONSTRAINT events_pkey PRIMARY KEY (sequenceno)
```
SandGrainOne, line 5: Is this table now using the same sequence as the app events table?

acn-sbuad: no, the contraint was renamed for events_app table

src/Events/Repository/CloudEventRepository.cs
```diff
@@ -49,6 +51,21 @@ public async Task CreateAppEvent(CloudEvent cloudEvent)
             pgcom.Parameters.Add(new NpgsqlParameter("cloudevent", cloudEvent.Serialize()) { Direction = System.Data.ParameterDirection.Input });
 
             await pgcom.ExecuteNonQueryAsync();
+
+            await using NpgsqlCommand pgcom2 = new NpgsqlCommand(string.Format(insertEventSql, serializedCloudEvent), conn);
```
SandGrainOne, line 55: We should avoid building a SQL statement with any form of string manipulation. Always use parameters. 

When it comes to parameters we've usually named them, but @vgr has found that using ordering is best. Not sure how much it matters though.

b-brodie-digdir: I find named parameters to be more readable and safer when refactoring. In this case, I would expect the performance difference should be non-measurable.

acn-sbuad: Given that it's only a single parameter and we have the full sql statement available I'm going with the numbered parameter here, the property name should be telling enough w.r.t readability / refactoring

src/Events/Repository/CloudEventRepository.cs
```diff
@@ -49,6 +51,33 @@ public async Task CreateAppEvent(CloudEvent cloudEvent)
             pgcom.Parameters.Add(new NpgsqlParameter("cloudevent", cloudEvent.Serialize()) { Direction = System.Data.ParameterDirection.Input });
 
             await pgcom.ExecuteNonQueryAsync();
+
+            await using NpgsqlCommand pgcom2 = new NpgsqlCommand(insertEventSql, conn)
+            {
+                Parameters =
```
SandGrainOne, line 57: Didn't know this syntax was an option. The Parameters property is read only. What is the result of this syntax? Is the new parameter added to existing collection or does it create a new collection?

SandGrainOne: Fun stuff:
https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQAYACdbAFgG4CD0BmMzEgYRIG8CTuzH0KSAWQAUASg5ceUlgHsANnICmAYwAuUGTAAqigM6rFAJxIH9RkgF4SMRQHcxnfFOc8AMlH2WSkl1Me+AkgAiU1Ugn0CAXwipSJonHmiE7gjYA0MYCDkmVnklNQ1tPXSJZOcGEnd9AB5yIgA+EgB9OQ9VeICKqtVa7AbKto4SAHNFdpIkzsZZBRV1TR0zQzFSwJ4WwasbWwGauvqxDt9JxIJIoA===

SandGrainOne: In other words. Elements are added

b-brodie-digdir (dismissed):

src/Events/Repository/CloudEventRepository.cs
```diff
@@ -35,11 +36,12 @@ public CloudEventRepository(IOptions<PostgreSQLSettings> postgresSettings)
         }
 
         /// <inheritdoc/>
-        public async Task CreateAppEvent(CloudEvent cloudEvent)
+        public async Task CreateAppEvent(CloudEvent cloudEvent, string serializedCloudEvent)
         {
             await using NpgsqlConnection conn = new NpgsqlConnection(_connectionString);
             await conn.OpenAsync();
 
+            await using var transaction = await conn.BeginTransactionAsync();
```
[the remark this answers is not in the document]
acn-sbuad: Might just be a question of notation here. The `using` without the curly braces means the block runs to the end of the function. So I believe this is already in place.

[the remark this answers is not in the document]
SandGrainOne: The transaction class does a rollback during dispose if not completed in some other way.
https://github.com/npgsql/npgsql/blob/docs/src/Npgsql/NpgsqlTransaction.cs/#L366

b-brodie-digdir (approved):

SandGrainOne (approved):

Issue #226, opened 2022-11-17
Remove obsolete procedures and functions from database


acn-sbuad:
### Description

There are some obsolete functions and procedures in the the events database. These should be deleted

### Tasks

- [ ] Set up migration script for deleting obsolete resources
- [ ] QA
- [ ] Manual test

### Acceptance Criterias

- [ ] Event flow is unchanged. 
  - [ ] Events are registered and persisted in DB
  - [ ] Events are successfully retrieved through app/party and app/ {org}/{app} endpoint
  - [ ] Events are successfully pushed to subscribers

PR #227, opened 2022-11-17
Remove unused functions and procedures

b-brodie-digdir:
Clean up unused functions and procedures now that we are fully deployed to flexible server.

See #226


acn-sbuad (approved):
🎈

Issue #109, opened 2022-09-09
Update the push outbound endpoint to handle generic events and app events


OddArneSaetervik:
### Description
The push endpoint currently responsible for identifying subscriptions and placing push messages in the events-outbound queue must be able to handle generic events.

Currently, if an event does not speficy a subject (an optional property for cloud event), we do not attempt to retrieve matching subscriptions. This is one of potentially more issues that need to be fixed to be able to handle push of genering events

### Tasks

- [x] Update GetSubscription function in database to handle all event properties being optional
- [x] Update OutboundService to handle all event properties that are optional
- [x] Introduce alternative authorization flow for generic events in AuthorizeAndEnqueueOutbound
- [x] QA
- [x] Manual test
~~- [ ] Implement k6 tests for new functionality ~~ handled in #166

### Acceptance Criterias
- [x] If an active subscriptions exists and event should be pushed to the endpoint
- [x] If an inactive subscriptions exists and event should be NOT be pushed to the endpoint
- [x] Regression test of push events for app events i successful

acn-sbuad:
Generic events missing properties _subject_ and _time_ are not pushed to subscribers

PR #276, opened 2023-02-07
Feature/109 push generic events

b-brodie-digdir:
Add support for generic events with missing subject field

## Related Issue(s)
- #109 

## Verification
- [x] All tests run green

## Documentation


src/Events/Services/OutboundService.cs
```diff
@@ -156,7 +160,7 @@ private Uri GetSourceFilter(Uri source)
             }
             else
             {
-                return null;
+                return source;
```
acn-sbuad, line 163: not sure why you're changing the return type here? Do we need to update the flow for app events?

src/Events/Migration/v0.24/03-setup-function.sql
```diff
@@ -0,0 +1,19 @@
...
+    RETURNS TABLE(id bigint, sourcefilter character varying, subjectfilter character varying, typefilter character varying, consumer character varying, endpointurl character varying, createdby character varying, validated boolean, "time" timestamp with time zone) 
+LANGUAGE 'plpgsql'
+AS $BODY$
+
+BEGIN
+return query 
+	SELECT s.id, s.sourcefilter, s.subjectfilter, s.typefilter, s.consumer, s.endpointurl, s.createdby, s.validated, s."time"
+	FROM events.subscription s
+  WHERE  s.sourcefilterhash = ANY(sourcehashset)
+  AND (s.subjectfilter is NULL OR length(subject) = 0 OR s.subjectfilter = subject)
```
acn-sbuad, line 14: do we need to check for null if we're always sending in an empty string from the repository service?

src/Events/Services/OutboundService.cs
b-brodie-digdir: GetSourceFilter is used as a cache key. I suggest refactoring to improve readability.

src/Events/Migration/v0.24/03-setup-function.sql
b-brodie-digdir: We don't always send in an empty string as far as I can tell. It should be possible to define a subject on a generic cloud event subscription. Will add a k6 test to verify that the logic is correct.

b-brodie-digdir: Reevaluated the insert_subscription codepath and concluded that the existing `events.getsubscriptions` implementation will continue to work now that we always send a non-null value for `subject`. 
Test cases where `subject` is empty string:
subscriptionfilter is null --> return
subscriptionfilter is empty string --> return
subscriptionfilter length > 0 --> do not return

Only thing I don't like is that we don't trim whitespace on string properties in our request models.

acn-sbuad (approved):
🎈

Issue #183, opened 2022-10-24
Add resource authorization to the POST events endpoint


acn-sbuad:
### Description

Initial implementation of events authorization will use hard coded attribute identificators to define correct resource in an authorization request. This is in line with how we currently do it for App events and subscriptions. The resource identifying values will be globally static. The resource registry might be needed in the future if there are authorization parameters that are different across different event producers.

Example of request is defined here: Examples for requests are specified here https://github.com/Altinn/altinn-authorization/issues/42

### Consideration
How do we enable applications to post events? 
Do we need to have strategies for when to require the events.publish scope? 
 - Will we have seperate endpoints but a common service for the rest of the flow? 
 
~~App events will not be handled by the new endpoint. The AccessToken used when authenticating an app will initially not be given the necessary scopes.~~

### Tasks

- [x] Specify resource attributes to specify a resource in an authorization request.
- [x] Implement authorization logic for external events
- [x] Implement unit tests
- [x] QA
- [x] Manual test
- [x] Automated tests	

### Acceptance Criteria

- [x] An end user should not be able to post an event for a source they are not registered to post for
- [x] Posting an event without the resource property should be prohibited
- [x] A request with the altinn:events.publish scope and authorized used should return 201
- [x] A request with the platform access token and an authorized used should return 201

acn-sbuad:
Ready for test. A significant change in this PR is the expected format of the cloud event. Resource should be on format: `urn:altinn:resource:ttd-altinn-events-automated-tests`

SandGrainOne:
The endpoint returns 200 OK and not 201 Created. With that being said we've agreed that 200 Ok is fine in this case. The event is accepted and stored, but we don't have any endpoint to retrieve the specific event. There is nowhere for a 201 Created responss to point to in a Location header.

SandGrainOne:
Removed acceptance criteria:
- [ ] An end user should not be able to post an event type not registered for the event source in the registry
It's no longer true.

SandGrainOne:
Manual testing completed

PR #318, opened 2023-03-31
Add authorization for POST events

acn-sbuad:
## Related Issue(s)
- #183

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Events/Authorization/PublishScopeOrAccessTokenHandler.cs
+++ b/src/Events/Authorization/PublishScopeOrAccessTokenHandler.cs
@@ -1,0 +1,51 @@
+﻿using System.Threading.Tasks;
+
+using Altinn.Platform.Events.Configuration;
+
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Altinn.Platform.Events.Authorization
+{
+    /// <summary>
+    /// Authorization handler for the <see cref="PublishScopeOrAccessTokenRequirement"/> class.
+    /// </summary>
+    public class PublishScopeOrAccessTokenHandler : AuthorizationHandler<PublishScopeOrAccessTokenRequirement>
+    {
+        private readonly IHttpContextAccessor _httpContextAccessor;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PublishScopeOrAccessTokenHandler"/> class.
+        /// </summary>
+        public PublishScopeOrAccessTokenHandler(IHttpContextAccessor httpContextAccessor)
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        /// <summary>
+        /// Handles verification of requirements. Publish scope requirement or AccesToken requirement must be satisfied.
+        /// </summary>
+        protected override async Task<object> HandleRequirementAsync(AuthorizationHandlerContext context, PublishScopeOrAccessTokenRequirement requirement)
+        {
+            var authorizationService = _httpContextAccessor.HttpContext.RequestServices.GetService<IAuthorizationService>();
+
+            var policyCheck = await authorizationService.AuthorizeAsync(context.User, AuthorizationConstants.POLICY_PLATFORM_ACCESS);
+
+            if (policyCheck.Succeeded)
+            {
+                context.Succeed(requirement);
+                return Task.CompletedTask;
+            }
+            
+            var scopeCheck = await authorizationService.AuthorizeAsync(context.User, AuthorizationConstants.POLICY_SCOPE_EVENTS_PUBLISH);
+
+            if (scopeCheck.Succeeded)
+            {
+                context.Succeed(requirement);
+            }
+            
+            return Task.CompletedTask;
+        }
+    }
+}
--- a/test/Altinn.Platform.Events.Tests/TestingControllers/EventsControllerTests.cs
+++ b/test/Altinn.Platform.Events.Tests/TestingControllers/EventsControllerTests.cs
@@ -467,7 +467,7 @@ namespace Altinn.Platform.Events.Tests.TestingControllers
             }
 
             [Fact]
-            public async Task Post_AuthorizedRequest_EventIsRegistered()
+            public async Task Post_PublishScopeIncluded_EventIsRegistered()
             {
                 // Arrange
                 string requestUri = $"{BasePath}/events";
@@ -490,6 +490,32 @@ namespace Altinn.Platform.Events.Tests.TestingControllers
                 // Assert
                 Assert.Equal(HttpStatusCode.OK, response.StatusCode);
             }
+
+            [Fact]
+            public async Task Post_PlatformAccessTokenIncluded_EventIsRegistered()
+            {
+                // Arrange
+                string requestUri = $"{BasePath}/events";
+
+                Mock<IEventsService> eventMock = new();
+                eventMock.Setup(em => em.RegisterNew(It.IsAny<CloudEvent>()))
+                          .ReturnsAsync(Guid.NewGuid().ToString());
+
+                HttpClient client = GetTestClient(eventMock.Object, true);
+                client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", PrincipalUtil.GetOrgToken("ttd"));
+
+                HttpRequestMessage httpRequestMessage = new(HttpMethod.Post, requestUri)
+                {
+                    Content = new StringContent(_validEvent.Serialize(), Encoding.UTF8, "application/cloudevents+json")
+                };
+                httpRequestMessage.Headers.Add("PlatformAccessToken", PrincipalUtil.GetAccessToken("ttd", "apps-test"));
+
+                // Act
+                HttpResponseMessage response = await client.SendAsync(httpRequestMessage);
+
+                // Assert
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            }            
 
             private HttpClient GetTestClient(IEventsService eventsService = null, bool enableExternalEvents = false)
             {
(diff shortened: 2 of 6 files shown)


src/Events/Authorization/PublishScopeOrAccessTokenHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Altinn.Platform.Events.Authorization
+{
+    /// <summary>
+    /// Authorization handler for the <see cref="PublishScopeOrAccessTokenRequirement"/> class.
+    /// </summary>
+    public class PublishScopeOrAccessTokenHandler : AuthorizationHandler<PublishScopeOrAccessTokenRequirement>
```
SandGrainOne, line 14: I would prefer these two policies to be separate. Is that causing issues?

SandGrainOne: Read a bit more about it. We must implement this as one policy, but it should be possible for one Policy to use two handlers. Might be something to look into?

acn-sbuad: Spent quite some time before I landed on this solution. Couldn't find anything else feasible to handle the OR condition. So I would need a concrete example or a commit to the PR to handle the implementation differently. 

That being said, there are two handlers being used. We are using the handler made available to the authorization service. Lines 33 and 41 call existing policies registered in Program.cs

acn-sbuad: I would suggest we keep it as is. Keeps us from adding an extra class and introducing more boilerplate code while also making it less obvious how the access is granted. One would need to check all handlers to see which requirement they are related to instead of it being 1:1 between requirement and handler.

src/Events/Authorization/PublishScopeOrAccessTokenHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        /// <summary>
+        /// Handles verification of requirements. Publish scope requirement or AccesToken requirement must be satisfied.
+        /// </summary>
+        protected override async Task<object> HandleRequirementAsync(AuthorizationHandlerContext context, PublishScopeOrAccessTokenRequirement requirement)
+        {
+            var authorizationService = _httpContextAccessor.HttpContext.RequestServices.GetService<IAuthorizationService>();
```
SandGrainOne, line 31: Could we try to depend on IAuthorizationService directly instead of going through the HttpContextAccessor?

acn-sbuad: Nope. it results in a circular dependency unfortunately..

acn-sbuad: "A circular dependency was detected for the service of type 'Microsoft.AspNetCore.Authorization.IAuthorizationService'.
Microsoft.AspNetCore.Authorization.IAuthorizationService(Microsoft.AspNetCore.Authorization.DefaultAuthorizationService) -> Microsoft.AspNetCore.Authorization.IAuthorizationHandlerProvider(Microsoft.AspNetCore.Authorization.DefaultAuthorizationHandlerProvider) -> System.Collections.Generic.IEnumerable<Microsoft.AspNetCore.Authorization.IAuthorizationHandler> -> Microsoft.AspNetCore.Authorization.IAuthorizationHandler(Altinn.Platform.Events.Authorization.AccessTokenHandler) -> Microsoft.AspNetCore.Authorization.IAuthorizationService"

SandGrainOne: hmm, okay. As we talked about. Fixing this probably requires changes in the AccessToken library. Maybe we should look into that?

SandGrainOne: Actually, what if the new requirement inherit from AccessTokenRequirement. That way AccessTokenHandler would accept the new requirement as input. Then make a new Handler for the scope check that use the same requirement.

Could that work?

SandGrainOne: I see one possible issue with my last idea. A requirement can't inherit from both ScopeAccessRequirement and AccessTokenRequirement. At least one of them would  need to be an interface.

test/Altinn.Platform.Events.Tests/TestingControllers/EventsControllerTests.cs
```diff
@@ -491,20 +492,87 @@ public async Task Post_AuthorizedRequest_EventIsRegistered()
...
+                Mock<IEventsService> eventMock = new();
+                eventMock.Setup(em => em.RegisterNew(It.IsAny<CloudEvent>()))
+                          .ReturnsAsync(Guid.NewGuid().ToString());
+
+                Mock<IAuthorization> authorizationMock = new();
+                authorizationMock.Setup(a => a.AuthorizePublishEvent(It.IsAny<CloudEvent>()))
+                            .ReturnsAsync(false);
+
+                HttpClient client = GetTestClient(eventMock.Object, authorizationMock.Object, true);
+                client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", PrincipalUtil.GetOrgToken("ttd"));
```
SandGrainOne, line 540: Should this have been a user token instead of an org token? Ref the test name.

acn-sbuad: Renamed test.

src/Events/Extensions/UriExtensions.cs
```diff
@@ -12,6 +12,8 @@ namespace Altinn.Platform.Events.Extensions
     /// </summary>
     public static class UriExtensions
     {
+        private static string _urnPattern = @"\burn:[a-z0-9][a-z0-9-]{0,31}:[a-z0-9()+,\-.:=@;$_!*'%/?#]+";
```
elsand, line 15: This will allow strings like `urn:1:::::""""` which we probably don't want.

```suggestion
        private static string _urnPattern = @"^urn:[a-z][a-z0-9-]{0,31}(:[a-z][a-z0-9()+,.\-=@;$_!*'%#]{0,99}){1,10}\/?$";
```
* Add `$` to disallow arbitrary trailing characters
* Replace `\b` with `^` to disallow whitespace at the beginning
* Require all "parts" to begin with a letter (ie. disallow "urn:1:2")
* Limit each "part" beyond the first to max 100 characters
* Disallow empty "parts" (ie. `urn:altinn:foo::::`)
* Require 1 to 10 parts beyond the first

acn-sbuad: Change seems to be breaking some of the unit tests. Will have a look

acn-sbuad: `urn:isbn:0451450523` is a valid urn ref https://en.wikipedia.org/wiki/Uniform_Resource_Name.
Rolling back to existing regex for now. Consider reworking the regex so existing unit tests run green @elsand ?

b-brodie-digdir: Could `System.Uri.IsWellFormedUriString`  with UriKind.Absolute and a check for scheme `urn` be an alternative?

https://learn.microsoft.com/en-us/dotnet/api/system.uri.iswellformeduristring?view=net-7.0

src/Events/Authorization/PublishScopeOrAccessTokenHandler.cs
b-brodie-digdir: @SandGrainOne did you find a way to use multiple handlers, like we looked at on Wednesday?

src/Events/Extensions/UriExtensions.cs
acn-sbuad: Did a quick test and `urn:altinn:foo:::::` passes the check for System.Uri.IsWellFormedUriString. I believe we verify that it is a well formed uri before we reach this point as well.

acn-sbuad: That being said.. if Microsoft deems it a well formed uri, we might be happy enough with that? Easier than maintaining the regex string at least.

elsand: `^urn:[a-z0-9-]{1,30}(:[a-z0-9()+,.\-=@;$_!*'%#\/]{1,100}){1,10}$` will allow numerics as the first (or sole) character in both NID and NSS parts. This also includes `/` as a valid NSS character, which seems reasonable (as we allow it as a trailing character)

We should perhaps consider restricting our support to a smaller subset of that standard ref wikipedia. I'm not sure it will ever make sense here to support percent-encoded arbitrary bytes in the NSS part, or query-components and fragments. I'm not too confident allowing eg. NUL bytes (even encoded) won't have any potential security implications down the line.

`^urn:[a-z0-9-]{1,30}(:[a-z0-9()+,.\-;$_!\/]{1,100}){1,10}$` disallows `% ? #` while still passing the unit tests and support a reasonably wide range of subdivisions in the NSS-es.

On the other hand, if we _are_ to support query-components and fragments, I think we should attempt to validate that those are well-formed (perhaps by explicitly calling `IsWellFormedUriString`)

b-brodie-digdir: If `System.Uri.IsWellFormedUriString` is _that_ lax, then for sure, being more restrictive sounds like a good idea. Could maybe add some test coverage for empty URN segments, to make sure we don't have unintended wildcard matching in the middle of a URN path.

src/Events/Authorization/PublishScopeOrAccessTokenHandler.cs
SandGrainOne: @b-brodie-digdir yes, we're currently testing changes in AccessTokenHandler and I've made a PR to update ScopeAccessHandler in the same way.

SandGrainOne (approved):

src/Events/Extensions/UriExtensions.cs
acn-sbuad: I think we can follow up on this at a later point.. Regardless the URN they choose for the resource will have to pass checks when being registered in the resource registry as well. We will sync with them on the validation.

Issue #305, opened 2023-03-13
Implement authorization of consumer for outbound events 


acn-sbuad:
### Description

Implementation of authorization has been left out of the outbound process. This should now be implemented.

Logic should be placed here: 
https://github.com/Altinn/altinn-events/blob/main/src/Events/Services/OutboundService.cs#L122

### Tasks
- [x] Extend GenericCloudEventXacmlMapper if needed
- [x] Make required changes to OutboundService
- [x] Implement unit tests
- [x] QA
- [x] Manual test
- [x] Automated tests (do we need these?)

### Acceptance Criterias
- [x] Generic events are successfully pushed to subscribers
- [x] If a subscriber is no longer authorized, the events should not be pushed (test by alter policy to limit access to the events to verify)

acn-sbuad:
Current status: Manual testing shows that authorization decision isn't permit. Looking further into this.

Issue #338, opened 2023-04-26
Modify posted events to include resource and resourceInstance


acn-sbuad:
## Description

We need to ensure that app events posted to the /app endpoint are enriched with resource and resource instance

## Task
- Extend logic in AppCloudEventExtensions.cs - CreateEvent
- Unit tests
- QA
- Manual test

## Acceptance criteria
- New cloud events that are registered from an app source include the resource and resource instance properties
- New cloud events that are returned from the GET endpoint include the resource and resource instance properties
- New cloude vents that are pushed to subscription endpoint include the resource and resource instance properties

acn-sbuad:
Verified in AT.

PR #339, opened 2023-04-26
Authorize consumers of generic events for outbound

acn-sbuad:
## Description

Added authorization of consumers for generic cloud events

## Related Issue(s)
- #305

## Verification
~~- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)~~ automated tests for events handled in another issue
- [x] All tests run green


src/Events/Extensions/CloudEventExtensions.cs
```diff
@@ -19,5 +20,26 @@ public static string Serialize(this CloudEvent cloudEvent, CloudEventFormatter f
...
+        /// Retrieves the resource instance from the cloud event. Returns null if it isn't defined.
+        /// </summary>
+        public static string GetResourceInstance(this CloudEvent cloudEvent) => cloudEvent["resourceinstance"]?.ToString();
+
+        /// <summary>
+        /// Sets the resource extension attribute of the cloud event
+        /// </summary>
+        public static void SetResource(this CloudEvent cloudEvent, string resource)
+        {
+            if (!string.IsNullOrEmpty(resource))
```
SandGrainOne, line 39: This if check should ideally be reflected in the method name. `SetResourceIfNull` or something.

acn-sbuad: We don't actually check that resource isn't defined on the cloud event, only that the input isn't empty. That being said, we probably should. Renaming and ensuring resource hasn't been defined before setting the value

src/Events/Services/OutboundService.cs
```diff
@@ -138,6 +146,16 @@ private static CloudEventEnvelope MapToEnvelope(CloudEvent cloudEvent, Subscript
             return cloudEventEnvelope;
         }
 
+        private static string GetAuthorizationCacheKey(string resource, string consumer)
+        {
+            if (resource == null)
```
SandGrainOne, line 151: Is there a chance for resource to be null? Would it even work with a cache key of null?

src/Events/Extensions/CloudEventExtensions.cs
SandGrainOne: oh, sorry, my comment was a mess.

My primary concern was the hidden if. Set methods should work like a setter and set the value. Think of how you would have handled this if Resource was a regular property.

acn-sbuad: Removed the hidden if check, and the method name reflects what it does. We don't override the resource name if it is already defined. Do you think we need an argument null exception here in addition to stop from setting an empty resource value?

src/Events/Services/OutboundService.cs
acn-sbuad: Maybe we could rename to setResourceIfNotDefined to be more precise?

src/Events/Extensions/CloudEventExtensions.cs
acn-sbuad: Maybe we could rename to setResourceIfNotDefined to be more precise?

SandGrainOne: The check should be done by the call point. Which I can't find in this PR. Where is the method being used?

acn-sbuad: Currently only used by tests. Was wondering why I implemented it at this point myself. But we will need it for #338 as well

src/Events/Services/OutboundService.cs
acn-sbuad: Not a real possibility, We ensure all events have a resource before they are placed in inbound queue. I'll remove the check.

src/Events/Extensions/CloudEventExtensions.cs
SandGrainOne: I'm happy with `SetResourceIfNotDefined`.

test/Altinn.Platform.Events.Tests/TestingExtensions/CloudEventExtensionsTests.cs
```diff
@@ -0,0 +1,133 @@
+﻿using System;
+
+using Altinn.Platform.Events.Extensions;
+
+using CloudNative.CloudEvents;
+
+using Xunit;
+
+namespace Altinn.Platform.Events.Tests.TestingExtensions
+{
```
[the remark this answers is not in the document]
SandGrainOne: I like the new file scoped namespace declaration.

SandGrainOne (approved):

test/Altinn.Platform.Events.Tests/TestingExtensions/CloudEventExtensionsTests.cs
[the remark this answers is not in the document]
acn-sbuad: For future reference, the commit suggestion must also remove a curly bracket at the end of the file to compile.

PR #343, opened 2023-05-09
added use case workflow for tt02

acn-sbuad:
- Use case workflow for TT02 including generic event cases
- fixed bug in subscription tests
- try/catch in all test files to ensure all errors not related to test results are identified


tba76 (approved):
🍺

PR #363, opened 2023-06-12
pushed content type is json, should accept this

acn-sbuad:
## Description
Test endpoint should accept application/json 

## Related Issue(s)
- #N/A


tba76 (approved):
🍺

PR #368, opened 2023-06-16
Increase query performance on GET

elsand:
This adds B-tree indices for queryable JSON columns and, adjusts function for improved query performance on generic events. 

## Description
The GIN index employed does not support  `cloudevent ->> 'id = ...` WHERE clauses, causing the `getevents` function to do sequential scans. This change:
* Drops the GIN index and adds B-tree indices for the columns that can be used for filtering.
* Replaces ILIKE with LIKE on the source-column, which helps index performance when querying (~20%). 
* Changes `VOLATILE PARALLEL UNSAFE` to `STABLE PARALLEL SAFE` to hint to Postgre that this is a pure function

Tests performed locally on 1.000.000 rows increases performance by an order of magnitude:

```sql
-- The table has 1.000.000 randomized entries, with a UUID after '/dialogs/', 
-- so a match for '/ab%' here will hit rows all over the table
SELECT * FROM events.getevents(
    _subject := NULL, 
    _alternativesubject := NULL, 
    _after := '0', 
    _type := NULL, 
    _source := 'https://dialogporten.no/api/v1/dialogs/ab%', 
    _size := 30
);
-- Query response times are reduced on my computer from ~500ms to ~50ms
```

Write performance appears to be slightly worse compared to the GIN index; a prepared INSERT statement called in a tight loop is reduced from ~1800 inserts/sec to ~1600 inserts/sec on my computer.

## Note! 
* The `CREATE INDEX` statements takes about ~90 seconds on my computer on 1.000.000 rows. 
* Changing `ILIKE` to `LIKE` might cause subscriptions/queries to no longer match anything

## Related Issue(s)
N/A

## Verification
- [ ] All tests run green

### DIFF
--- a/src/Events/Migration/v0.32/02-update-function.sql
+++ b/src/Events/Migration/v0.32/02-update-function.sql
@@ -58,7 +58,6 @@ CREATE OR REPLACE FUNCTION events.getevents(
 	_after character varying,
 	_type text[],
 	_source character varying,
-	_resource character varying,
 	_size integer)
     RETURNS TABLE(cloudevents text)
     LANGUAGE 'plpgsql'
--- a/src/Events/Migration/v0.32/02-update-function.sql
+++ b/src/Events/Migration/v0.32/02-update-function.sql
@@ -1,3 +1,53 @@
+DROP FUNCTION IF EXISTS events.getappevents(character varying, character varying, timestamp with time zone, timestamp with time zone, text[], text[], integer);
+CREATE OR REPLACE FUNCTION events.getappevents(
+	_subject character varying,
+	_after character varying,
+	_from timestamp with time zone,
+	_to timestamp with time zone,
+	_type text[],
+	_source text[],
+	_size integer)
+    RETURNS TABLE(cloudevents text)
+    LANGUAGE 'plpgsql'
+    --COST 100
+    --VOLATILE PARALLEL UNSAFE
+    ROWS 1000
+
+AS $BODY$
+DECLARE
+_sequenceno bigint;
+BEGIN
+IF _after IS NOT NULL AND _after <> '' THEN
+	SELECT
+		case count(*)
+		when 0
+			then 0
+		else
+			(SELECT sequenceno FROM events.events
+			WHERE cloudevent->>'id' = _after
+			ORDER BY sequenceno ASC)
+		end
+	INTO _sequenceno
+	FROM events.events
+	WHERE cloudevent->>'id' = _after;
+END IF;
+return query
+	SELECT cast(cloudevent as text) as cloudevents
+		FROM events.events
+		WHERE (_subject IS NULL OR cloudevent->>'subject' = _subject)
+			AND (_from IS NULL OR cloudevent->>'time' >= _from::text)
+			AND (_to IS NULL OR cloudevent->>'time' <= _to::text)
+			AND registeredtime <= now() - interval '30 second'
+			AND (_type IS NULL OR cloudevent->>'type' ILIKE ANY(_type))
+			AND (_source IS NULL OR cloudevent->>'source' ILIKE ANY(_source))
+			AND (_after IS NULL OR _after = '' OR sequenceno > _sequenceno)
+		ORDER BY sequenceno
+		limit _size;
+END;
+$BODY$;
+
+
+
 -- Changes:
 --  * Sets to STABLE PARALLEL SAFE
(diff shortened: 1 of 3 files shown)


src/Events/Migration/v0.32/02-update-function.sql
```diff
@@ -0,0 +1,99 @@
...
+--  * Sets to STABLE PARALLEL SAFE
+--  * Changes ILIKE to LIKE for better index performance
+
+CREATE OR REPLACE FUNCTION events.getevents(
+	_subject character varying,
+	_alternativesubject character varying,
+	_after character varying,
+	_type text[],
+	_source character varying,
+	_resource character varying,
```
elsand, line 61: @HenningNormann Remove _resource for this PR?

src/Events/Migration/v0.32/02-update-function.sql
```diff
@@ -0,0 +1,99 @@
...
+		ORDER BY sequenceno
+		limit _size;
+END;
+$BODY$;
+
+
+
+-- Changes:
+--  * Sets to STABLE PARALLEL SAFE
+--  * Changes ILIKE to LIKE for better index performance
```
acn-sbuad, line 53: if changing to LIKE we should ensure that all data persisted in db currently is lower case as well as ensure that we change all input to lower case before it reached this db function.

src/Events/Migration/v0.32/02-update-function.sql
HenningNormann: Yes

tba76: If not this will be a new function, not replacing the old one.

src/Events/Migration/v0.32/02-update-function.sql
elsand: @HenningNormann 
Re-ran the tests now, and it appears to have very significant effect on my computer (much more than indicated in the PR description). I have:
* 1,200,000 rows with synthetized data (not totally random, see the gist linked below). Source has a common prefix and totally random suffix (like a typical instance URL would have, ie. `https://dialogporten.no/api/v1/dialogs/bba8b10a-3151-490d-8a58-30d764e269d`)
* Query that filters on `_source='https://dialogporten.no/api/v1/dialogs/${TWO_RANDOM_09AZ_CHARS}%'`, `_size=30`, `_after='0'`, rest `NULL`

This gives on my computer  ~9500 qps with `LIKE` and ~450 qps with `ILIKE` (using Postgre 13). Dropping the random hex from the query, ie using `https://dialogporten.no/api/v1/dialogs/%`, performs unsurprisingly much better, but still with `LIKE` being ~3x faster (~60000 qps and 20000 qps for `LIKE` and `ILIKE`, respectively.)

See [this gist](https://gist.github.com/elsand/c5b86ef1af894028700fd5601fbd2ffa) for the code I use to test both insert and select performance. Can you check to see if you get similar results?

elsand: Updated the gist slightly, now does 5 runs of 10 seconds (my laptop has lots of background noise and performance sometimes varies greatly):
```
# using ILIKE
PS C:\repos\populate-eventsdb\populate-eventsdb> dotnet run -- select 
Run #1: time: 10351ms, selects: 5190, avg: 501,4008308376002qps
Run #2: time: 10308ms, selects: 4560, avg: 442,37485448195577qps
Run #3: time: 10350ms, selects: 4350, avg: 420,2898550724638qps
Run #4: time: 10470ms, selects: 4650, avg: 444,12607449856733qps
Run #5: time: 10410ms, selects: 4770, avg: 458,21325648414984qps

-----
Total number of selects: 23520
Average qps: 453,27526065254676
----

# using LIKE
PS C:\repos\populate-eventsdb\populate-eventsdb> dotnet run -- select
Run #1: time: 10018ms, selects: 109500, avg: 10930,325414254341qps
Run #2: time: 10023ms, selects: 109470, avg: 10921,87967674349qps
Run #3: time: 10018ms, selects: 103860, avg: 10367,338790177679qps
Run #4: time: 10026ms, selects: 103800, avg: 10353,081986834231qps
Run #5: time: 10024ms, selects: 95100, avg: 9487,230646448525qps

-----
Total number of selects: 521730
Average qps: 10411,90205352332
----
```

acn-sbuad (approved):
🎈

PR #372, opened 2023-06-20
set resource and resourceinstance for app events

acn-sbuad:
## Description
Mapping from app event to generic event now includes setting extension attributes resource and resourceinstance. 

## Related Issue(s)
- #338

## Verification
- [x] All tests run green

### DIFF
--- a/src/Events/Configuration/AuthorizationConstants.cs
+++ b/src/Events/Configuration/AuthorizationConstants.cs
@@ -33,6 +33,11 @@
         /// <summary>
         /// Scope for allowing subscribing to events
         /// </summary>
-        public const string SCOPE_EVENTS_SUBSCRIBE = "altinn:events.subscribe";        
+        public const string SCOPE_EVENTS_SUBSCRIBE = "altinn:events.subscribe";
+
+        /// <summary>
+        /// The urn prefix for Altinn App resources
+        /// </summary>
+        public const string AppResourcePrefix = "urn:altinn:resource:altinnapp.";
     }
 }
--- a/src/Events/Extensions/AppCloudEventExtensions.cs
+++ b/src/Events/Extensions/AppCloudEventExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 
+using Altinn.Platform.Events.Configuration;
 using Altinn.Platform.Events.Models;
 
 using CloudNative.CloudEvents;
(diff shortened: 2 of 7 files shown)


src/Events/Extensions/AppCloudEventExtensions.cs
```diff
@@ -26,12 +30,43 @@ public static CloudEvent CreateEvent(this AppCloudEventRequestModel appEvent)
...
             return cloudEvent;
         }
+
+        /// <summary>
+        /// Decomposes the source of a cloud event into known parts.
+        /// </summary>
+        /// <param name="source">The uri source of the cloud event</param>
+        /// <returns>A tuple containing org, app, intanceOwnerPartyId and instanceGuid</returns>
+        /// <remarks>Domain should always be validated as a match to current environment before using this method.</remarks>
+        public static (string Org, string App, string InstanceOwnerPartyId, string InstanceGuid) GetPropertiesFromAppSource(Uri source)
```
SandGrainOne, line 52: I assume the need for this method is temporary? I don't like where we added it, but I can close my eyes for things that are actually temporary. Will we be able to remove it again when all apps are producing events with resource and resourceinstance properties?

tba76 (approved):
🍺

src/Events/Extensions/AppCloudEventExtensions.cs
acn-sbuad: I don't think this will be removed any time soon as authorization of app events will still require these parameters. I think the agreement with team authorization was that we would provide these attributed in the decision request rather than them having to pick them out on their end.

acn-sbuad: Yes we should be able to remove it once apps start to produce these properties themselves. We will still need some decomposition of the resource and resourceinstance to correcly populate the decision request with attributes, but that would then be limited to the xacml mapper.

src/Events/Configuration/AuthorizationConstants.cs
```diff
@@ -33,6 +33,11 @@ public static class AuthorizationConstants
         /// <summary>
         /// Scope for allowing subscribing to events
         /// </summary>
-        public const string SCOPE_EVENTS_SUBSCRIBE = "altinn:events.subscribe";        
+        public const string SCOPE_EVENTS_SUBSCRIBE = "altinn:events.subscribe";
+
+        /// <summary>
+        /// The urn prefix for Altinn App resources
+        /// </summary>
+        public const string AppResourcePrefix = "urn:altinn:resource:altinnapp.";
```
b-brodie-digdir, line 41: Would "urn:altinn:resource:app." also work?

src/Events/Extensions/AppCloudEventExtensions.cs
b-brodie-digdir: do we have a good way of making sure that org and app names don't contain delimiters that can trip us up when doing this kind of decomposition? I'm guessing that both "." and ":" could cause problems.

PR #375, opened 2023-06-22
Add administrative publishing scope short circuiting authorization

elsand:
This adds a way to authorize publishing of all event types if a special scope is provided in addition to the ordinary authorization policy.

## Description
This adds a short circuit in the AuthorizationService, checking if the authenticated and authorized principal has supplied a scope-claim containing the `altinn:events.publish.admin`, which will the cause the HTTP request to PEP to be omitted. This will be used by Dialogporten when producing events on behalf of service owners on with arbitrary `resource`-properties. This avoids having to add a rule in all XACMLs for "publish" to Digdir, and avoids the external HTTP requests against Authorization which should help performance.

This also fixes `HasRequiredScope(this ClaimsPrincipal user, string requiredScope)` so that it splits on space and uses `Equals` instead of `Contains` to avoid incorrectly matching on `unrelatedprefix:altinn:events...`

## Related Issue(s)
- https://github.com/digdir/dialogporten/issues/52

## Verification
- [x] All tests run green

### DIFF
--- a/src/Events/Configuration/AuthorizationConstants.cs
+++ b/src/Events/Configuration/AuthorizationConstants.cs
@@ -36,6 +36,11 @@
         public const string SCOPE_EVENTS_SUBSCRIBE = "altinn:events.subscribe";
 
         /// <summary>
+        /// Scope for allowing administrators to publish events without Altinn Authorization
+        /// </summary>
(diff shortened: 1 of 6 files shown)


tba76 (approved):
🍺

PR #376, opened 2023-06-23
Initial poc for resource attribute matching of cloud event subjects

elsand:
This implements a approach to support role/accessgroup-based authorization of service resources refered to by generic cloud events.

## Description
This simply extends `CreateResourceCategory` used by `GenericCloudEventMapper.CreateDecisionRequest` to parse the subject-property of the given cloudevent, and - if it matches a known prefix - maps these attributes to the resource category. This will allow the PDP to enrich the subject with roles/access groups, and apply to POST and (current) GET actions for the events-API, outbound pushed events and presumably subscription requests (once implemented)

### Considerations
* There are no attempts made to ensure that the XACML subject attributes contains something that the PDP is able to use in order to get roles/access groups. With Altinn-tokens, the userId claim is present and will be copied into the authorization request. 
* The event component does not attempt to resolve the subject to partyId, this should be the PDPs job to do  (and cache) - and can determine to not do so if the relevant policy does not require it
* The PDP currently supports only [partyId and organization numbers](https://github.com/Altinn/altinn-authorization/blob/main/src/Authorization/Services/Implementation/ContextHandler.cs#L167), not SSNs, but this can be easily extended.
* The GET endpoint can at some point be optimized to perform an authorization request based on the supplied service resource first, and use obligations from the PEP to filter subjects/types in the database query.

### DIFF
--- a/src/Events/Authorization/GenericCloudEventXacmlMapper.cs
+++ b/src/Events/Authorization/GenericCloudEventXacmlMapper.cs
@@ -15,6 +15,10 @@ namespace Altinn.Platform.Events.Authorization
     /// </summary>
     public static class GenericCloudEventXacmlMapper
     {
+        private const string SubjectTypeOrg = "org";
+        private const string SubjectTypePerson = "person";
+        private const string SubjectTypeParty = "party";
+
         /// <summary>
         /// Create XACML request for authorizing the provided action type on multiple events
         /// </summary>
@@ -113,6 +117,8 @@ namespace Altinn.Platform.Events.Authorization
                 resourceCategory.Attribute.Add(DecisionHelper.CreateXacmlJsonAttribute(AltinnXacmlUrns.ResourceInstance, cloudEvent["resourceinstance"].ToString(), defaultType, defaultIssuer));
             }
 
+            AddResourceReporteeAttributeFromCloudEventSubject(cloudEvent, resourceCategory);
+
             return resourceCategory;
         }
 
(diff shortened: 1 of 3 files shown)


src/Events/Authorization/XacmlMapperHelper.cs
```diff
@@ -25,6 +25,8 @@ public static class XacmlMapperHelper
 
     // urn:altinn:organization:identifier-no is a value defined by Authorization so we need to use 'z' here.
```
acn-sbuad, line 26: This comment on line 26 can be removed. We have concluded that 'z' is the way to go, no reason to explain why we use a 'z' notation moving forward.

src/Events/Authorization/XacmlMapperHelper.cs
```diff
@@ -79,4 +81,41 @@ public static (string AttributeId, string AttributeValue) SplitResourceInTwoPart
...
+    /// or "urn:altinn:person:identifier-no", which is copied to the resource attribute with the corresponding URN.
+    /// This will enable the PDP to enrich the request with roles etc. that the user has in the context of the CloudEvent
+    /// subject (aka reportee).
+    ///
+    /// Also note that this requires a XACML subject attribute that the PDP understands in order to look up the user's
+    /// roles/access groups for that particular reportee, eg. "urn:altinn:userid" or "urn:altinn:person:identifier-no".
+    /// </remarks>
+    /// <param name="cloudEvent">CloudEvent containing a subject representing a party (ie. reportee)</param>
+    /// <param name="resourceCategory">The XACML resource category to be populated</param>
+    public static void AddResourceReporteeAttributeFromCloudEventSubject(CloudEvent cloudEvent, XacmlJsonCategory resourceCategory)
```
acn-sbuad, line 99: Can we consider renaming the function? The term reportee is not much used in the platform components we are responsible for. 
```suggestion
    public static void AddResourceSubjectReporteeAttributeFromCloudEvent(CloudEvent cloudEvent, XacmlJsonCategory resourceCategory)
```

elsand: Changed to `AddResourcePartyAttributeFromCloudEventSubject`

acn-sbuad (approved):
🎈

Issue #316, opened 2023-03-28
Extend GET endpoint to include resource as query param


acn-sbuad:
### Description

Two new properties: resource and resourceId have been defined for cloudEvents processed through Altinn. 
Event subscribers should be able to request a list of events based on the resource attribute. 

### Consideration
- Do we need to use source as a query parameter once resource is required? 
  - No, support for source, will be removed as query param. 
- Consider if we need to check if after is defined in function

### Tasks
- [x] Add new required query parameter to endpoint
- [x] Remove support for source as query parameter
- [x] Update function in postgreSQL to match on resource and remove support for source
- [x] Update function to use optimal queries for jsonb 
- [x] QA
- [x] Update swagger
- [x] Test
- [x] [Documentation](https://github.com/Altinn/altinn-studio-docs/pull/1167)

### Acceptance Criteria
- [x] It is possible to query events using the resource query parameter
- [x] A request without resource parameter should fail with bad request
- [x] A request with resource and after parameter is valid
- [x] Indexes are utilized when getevents function is called.
- [x] ~~Querying app events should also utilize resource, if not possible to consolodate procedures ensure seperate issues is created~~ Moved to seperate issue

SandGrainOne:
There are no index on resource.

PR #403, opened 2023-09-01
resource as query param generic endpoint

acn-sbuad:
## Description

- Check if indexes are used
- Optimize jsonb queries 

## Related Issue(s)
- #316 

## Verification
- [x] Relevant automated test added:PR #404 
- [x] All tests run green


src/Events/Controllers/EventsController.cs
```diff
@@ -99,8 +100,8 @@ public async Task<ActionResult<string>> Post([FromBody] CloudEvent cloudEvent)
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
         [Produces("application/cloudevents+json")]
         public async Task<ActionResult<List<CloudEvent>>> Get(
+            [FromQuery, Required] string resource,
```
SandGrainOne, line 103: What happens if the parameter is missing? Will the response be 400 Bad Request or will it be 404 Not Found?

src/Events/Swagger/RequestBodyExampleFilter.cs
```diff
@@ -41,6 +41,8 @@ private static void CreateCloudEventRequestModelExamples(OpenApiRequestBody requ
                 ("Instance created event with alternative subject",
                 CreateOpenApiObject(new List<(string Name, string Value)>()
                 {
+                    ("resource", "urn:altinn:resource:altinnapp.ttd.apps-test"),
```
SandGrainOne, line 44: Should we update the example for the Generic POST event endpoint as well?

acn-sbuad: the generic post example is based on the cloud event schema. Added resource prop to that schema as well. We don't have a set of example request bodies as we have for other endpoints.

src/Events/Controllers/EventsController.cs
SandGrainOne: Good

SandGrainOne (approved):

PR #413, opened 2023-09-08
Add index for cloudevent.resource

SandGrainOne:
## Description
Adding an index on the cloudevent->>'resource' "column" in the events table.

## Related Issue(s)
- #316 

## Verification
- [x] All tests run green

## Documentation


HenningNormann:
I don't think this index will ever be used, because it's not very selective. Is there a change in the pipeline to ensure that the source field is used in most use cases? I suggest that we evaluate the index usage when such a change is in production.

HenningNormann:
Ref. slack call. I suggest you change to:

  CREATE INDEX IF NOT EXISTS idx_events_cloudevent_resource_sequenceno
    ON events.events ((cloudevent ->> 'resource'), sequenceno);

This index was created in 1 second in yt01, and it was used (proved by auto_explain) for a typical query with resource, after ("low" value to trigger worst case) and size.

HenningNormann (approved):

PR #427, opened 2023-10-31
Replace Altinn3.ruleset with .editorconfig

SandGrainOne:
## Description
Replacing the deprecated Altinn3.ruleset file with a .editorconfig file with the same set of rules. Making sure all projects are updated. Fixing errors and warnings


acn-sbuad (approved):
🎈

PR #461, opened 2024-01-24
Rollback dependencies

SandGrainOne:
## Description
The updated packages doesn't work. Need to do a more systematic test.


acn-sbuad (approved):
🎈

Issue #469, opened 2024-02-08
Authorization in generic events needs to support a consumer of type organisation number


SandGrainOne:
### Description

Authorization of generic events is currently being done based on a party id. This works fine if the policy for the resource contains party id for the subjects with access. Unfortunately it is impractical for the people creating those policies to obtain correct party id for every organization.

We should be able to authorize access to generic events for resources with rules that lists specific organisation numbers in their policy.

acn-sbuad:
Push to organization consumer verified in at23.

PR #477, opened 2024-02-16
Changes to subject indicators

SandGrainOne:
## Description
Making multiple changes to how we handle subject prefixes and values:
- CreatedBy and Consumer
    - Added support for `/organisation/` in access subject authorization. 
    - Storing new subscriptions for an organisation as `/organisation/` instead of `/party/`. One thing to note is that there are no subscription in any environment where CreatedBy or Consumer data have the `/party/` prefix. They are all either `/org/` or `/user/`. This means that at no point in time has anyone successfully created an organisation subscription.
- Alternative subject
    - Added support for alternative subject using `/organisation/` as prefix. Keeping `/org/` for backwards compatibility. This might be in use by external parties when setting subject filter on a subscription.

Other changes:
- `SubscriptionService.GetAllSubscriptions()` where the only method taking consumer as input. Changed to use `IClaimPrincipalProvider` instead.
- Removed support for Hungarian prefixes on variable names

## Related Issue(s)
- #469 

## Verification
- [x] All tests run green

## Documentation


src/Events/Authorization/XacmlMapperHelper.cs
```diff
@@ -17,15 +17,17 @@ public static class XacmlMapperHelper
     private const string UserPrefix = "/user/";
     private const string OrgPrefix = "/org/";
     private const string PartyPrefix = "/party/";
-    private const string OrganizationPrefix = "/organization/";
+    private const string OrganisationPrefix = "/organisation/";
```
acn-sbuad, line 20: shouldn't this also be 'z'  ? ref comment on line 27.

SandGrainOne: The comment is for a specific value used by the Authorization logic. I would have used `s` if I could.
I'll try to make it more obvious what I ment.

src/Events/Services/AppSubscriptionService.cs
```diff
@@ -137,16 +133,26 @@ private async Task<string> GetPartyFromAlternativeSubject(string alternativeSubj
         {
             int partyId = 0;
 
-            if (alternativeSubject.StartsWith(OrganisationPrefix))
+            /* Both OrgPrefix and OrganisationPrefix is assumed to give us the organisation number of an 
+             * organisation. OrgPrefix is NOT for the application owner acronym like it is many other 
+             * places. OrgPrefix was kept as is when adding OrganisationPrefix for backwards compability.
+             */
+
+            if (alternativeSubject.StartsWith(OrgPrefix))
```
acn-sbuad, line 141: Can we consider handling the orgPrefix and OrganisationPrefix in the same if-clause to better demonstrate that they represent the same "thing" ?

acn-sbuad: Alternatively, we could have the controller be responsible for mapping the alternative subject to a format that the service can handle. E.g. mapping all /org/ to /organisation/. 
Limiting the backward compatibility to the controller and making it clear there that use of /org/ is deprecated?

SandGrainOne: I like the code flow the way it is, but I moved the to cases closer together and improved the explanation.

acn-sbuad (approved):
🎈

PR #482, opened 2024-02-22
breaking changes labeler workflow

acn-sbuad:
## Description
Labeler workflow failing since upgrade to v5 of action


.github/workflows/pr-labeler.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - uses: actions/checkout@v4
```
SandGrainOne, line 17: This is adding checkout, but not changing labeler.

acn-sbuad: only the config requires changing. v5 of labeler includes the breaking changes, but config was not updates when action was patched

SandGrainOne: Okay, so the checkout step doesn't require any input?

acn-sbuad: no

SandGrainOne (approved):

PR #527, opened 2024-04-26
Use appsettings.unittest.json in unit tests

HenningNormann:
## Description
Renamed appsettings.json to appsettings.unittest.json in unit tests and overrided configuration to use this file in unit tests to avoid conflict between the unit test config and the controllers config

## Related Issue(s)
- #527


acn-sbuad (approved):
🎈

PR #541, opened 2024-06-13
Automated test workflows added

acn-sbuad:
Added regression test workflow for all environments
Added use case workflows for production

### DIFF
--- a/.github/workflows/regression-test-TT02.yaml
+++ b/.github/workflows/regression-test-TT02.yaml
@@ -11,24 +11,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Run app-events use case tests
+    - name: Run app-events regression tests
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/app-events.js
         flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }} -e app=apps-test -e userName=${{ secrets.USER_NAME }} -e userPassword=${{ secrets.USER_PASSWORD }}
-    - name: Run events use case tests (POST)
+    - name: Run events regression tests (POST)
       if: always()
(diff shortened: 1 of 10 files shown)


.github/workflows/regression-test-TT02.yaml
```diff
@@ -0,0 +1,54 @@
...
+  workflow_dispatch:
+  schedule:
+  - cron: '0 12 * * 1-5'
+
+jobs:
+  regression-tests:
+    environment: TT02
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
```
SandGrainOne, line 13: Is this line it's own step without a name?

acn-sbuad: Correct. Usually don't name this simple step because the action itself is pretty clear and we don't have any additional parameters to add either, so would just be increasing the number of lines without adding much value

acn-sbuad: image
 this is what it looks like during a workflow run

SandGrainOne: Okay, thank you.

.github/workflows/regression-test-TT02.yaml
```diff
@@ -0,0 +1,54 @@
...
+        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }} -e app=apps-test -e webhookEndpoint=${{ secrets.WEBHOOK_ENDPOINT }}
+    - name: Build failure report
+      if: failure()
+      run: |
+          report=":warning: Events regression test failure in ${{ vars.ENV }} :warning: \n"
+          report+="\n Workflow available here: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          echo "stepreport="$report >> $GITHUB_ENV
+    - name: Report failure to Slack
+      if: failure()
+      id: slack
```
SandGrainOne, line 45: This `id` property is included only for the slack action. Is it a requirement?

SandGrainOne (approved):

Issue #596, opened 2024-10-02
Events - Implement support for system user


olebhansen:
```[tasklist]
### Blockers
- [ ] https://github.com/Altinn/altinn-authorization/issues/578
```

**Observations:**

- [ ] The `PrincipalUtil` class provides several methods for creating instances of `ClaimsPrincipal` and tokens., primarily for testing purposes. Introducing a system-user model specifically designed for testing could enhance both consistency and maintainability, especially when writing tests for system-user scenarios.

- [x] The `ClaimsPrincipalExtensions` class provides several extension methods built on the `ClaimPrincipal` type. A thorough review of these extensions is necessary to ensure they function correctly and as intended across their current usage scenarios.

- [x] The `ClaimsPrincipalProvider` implements the `IClaimsPrincipalProvider` and includes a `GetUser` function utilized by several components, such as `SubscriptionService`, `SubscriptionServiceTest`, `AuthorizationService`, `AuthorizationServiceTest`, `AppSubscriptionService`, `AppSubscriptionServiceTest`, `GenericSubscriptionService`, `GenericSubscriptionServiceTest`, and `OutboundServiceTests`. A review of this function, along with the `JwtTokenMock` class, is essential to ensure its compatibility with the system-user.

- [x] The `CloudEventXacmlMapper` class is responsible for converting events into XACML requests. It includes a method for creating XACML requests for multiple resources and actions using the claims principal user. This class, along with the `AppCloudEventXacmlMapper` and `AppCloudEventXacmlMapperTests` classes, should be carefully reviewed to ensure compatibility with the system-user.

- [x] The `PepWithPDPAuthorizationMockSI` class, which simulates a Policy Decision Point (PDP) in an authorization system, evaluates access control policies to make authorization decisions. A thorough review is needed to confirm its alignment with the system-user.

- [x] The `IdentityTelemetryFilter` class reads data from `ClaimsPrincipal` instances. Its logic may need to be added or updated to ensure proper support for the system-user.

- [ ] The `GenericCloudEventXacmlMapper` class, designed to convert cloud events into XACML (eXtensible Access Control Markup Language) requests for authorization within the Altinn Platform, provides methods to create XACML requests for single and multiple cloud events. We may need to update its logic to accommodate the system-user.

[...]

PR #652, opened 2025-01-27
Integrate the system-user in the Events API

Ahmed-Ghanam:
## Description
Integrate the system-user in the Events API.

## Related Issue(s)
- #596 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Events/Authorization/AppCloudEventXacmlMapper.cs
+++ b/src/Events/Authorization/AppCloudEventXacmlMapper.cs
@@ -9,145 +9,144 @@ using Altinn.Platform.Events.Extensions;
 
 using CloudNative.CloudEvents;
 
-namespace Altinn.Platform.Events.Authorization
+namespace Altinn.Platform.Events.Authorization;
+
+/// <summary>
+/// Utility class for converting Events to XACML request.
+/// </summary>
+public static class AppCloudEventXacmlMapper
 {
     /// <summary>
-    /// Utility class for converting Events to XACML request.
+    /// Creates an XACML request for multiple events using the claims principal user as the subject.
     /// </summary>
-    public static class AppCloudEventXacmlMapper
+    /// <param name="user">The claims principal user.</param>
+    /// <param name="events">The list of events.</param>
+    /// <returns>A <see cref="XacmlJsonRequestRoot"/> object representing the XACML request.</returns>
+    public static XacmlJsonRequestRoot CreateMultiDecisionReadRequest(ClaimsPrincipal user, List<CloudEvent> events)
     {
-        /// <summary>
-        /// Creates an XACML request for multiple events using the claims principal user as the subject.
-        /// </summary>
-        /// <param name="user">The claims principal user.</param>
-        /// <param name="events">The list of events.</param>
-        /// <returns>A <see cref="XacmlJsonRequestRoot"/> object representing the XACML request.</returns>
-        public static XacmlJsonRequestRoot CreateMultiDecisionReadRequest(ClaimsPrincipal user, List<CloudEvent> events)
+        List<string> actionTypes = ["read"];
+        var resourceCategories = CreateMultipleResourceCategory(events);
+        return CloudEventXacmlMapper.CreateMultiDecisionRequest(user, actionTypes, resourceCategories);
+    }
+
+    /// <summary>
+    /// Creates a decision request based on a cloud event and subject.
+    /// </summary>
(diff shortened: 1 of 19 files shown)


src/Events/Authorization/AppCloudEventXacmlMapper.cs
```diff
@@ -9,129 +9,144 @@
...
-            List<string> actionTypes = new() { "read" };
+            Action = [],
+            Resource = [],
+            AccessSubject = [],
+        };
 
-            var resourceCategory = CreateMultipleResourceCategory(events);
-            return CloudEventXacmlMapper.CreateMultiDecisionRequest(user, actionTypes, resourceCategory);
-        }
+        (string orgId, string appId, string instanceOwnerPartyId, string instanceGuid) = AppCloudEventExtensions.GetPropertiesFromAppSource(cloudEvent.Source);
```
SandGrainOne, line 47: If you're unhappy with the name `org` we can change it to `applicationOwnerId`.

`appId` should be `appName`. `appId` would be the complete identifier for an app that contains both `applicationOwnerId` and `appName` separated by a `/`.

SandGrainOne: I understand it can be confusing when the PEP library calls them something else.

src/Events/Services/SubscriptionService.cs
```diff
@@ -8,150 +9,156 @@
...
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="SubscriptionService"/> class.
-        /// </summary>
-        public SubscriptionService(
-            ISubscriptionRepository repository,
-            IAuthorization authorization,
-            IEventsQueueClient queue,
-            IClaimsPrincipalProvider claimsPrincipalProvider)
+        if (!await _authorization.AuthorizeConsumerForEventsSubscription(eventsSubscription))
```
SandGrainOne, line 48: I was wrong when we talked about this. We do need to update the `XacmlMapperHelper.AddSubjectAttribute()` method. It's used in a call chain (here) where subject is the event subscriber. I thought it was limited to when subject is the "instance owner" of an instance.

The term Subject is used in two different contexts.
1. The active user trying to access some data/event.
2. The subject of an event.

src/Events/Authorization/AppCloudEventXacmlMapper.cs
Ahmed-Ghanam: I have made the necessary updates.

src/Events/Authorization/XacmlMapperHelper.cs
```diff
@@ -19,14 +19,16 @@ public static class XacmlMapperHelper
     private const string OrgPrefix = "/org/";
     private const string PartyPrefix = "/party/";
     private const string OrganisationPrefix = "/organisation/";
+    private const string SystemUserPrefix = "/systemuser/";
 
     private const string ClaimUserId = "urn:altinn:userid";
     private const string ClaimOrg = "urn:altinn:org";
     private const string ClaimPartyID = "urn:altinn:partyid";
+    private const string ClaimSystemUserId = "urn:altinn:systemuser";
```
SandGrainOne, line 27: Correct value is: `urn:altinn:systemuser:uuid`

It's actually defined in the [PEP package](https://github.com/Altinn/altinn-authorization/blob/main/src/Altinn.Common.PEP/Altinn.Common.PEP/Constants/AltinnXacmlUrns.cs).

test/Altinn.Platform.Events.Tests/Utils/PrincipalUtil.cs
```diff
@@ -1,126 +1,221 @@
...
-        }
+        return JwtTokenMock.GenerateToken(principal, new TimeSpan(0, 1, 5));
+    }
+
+    /// <summary>
+    /// Creates a ClaimsPrincipal object representing a system user.
+    /// </summary>
+    /// <param name="systemId">The system identifier.</param>
+    /// <param name="systemUserId">The system user identifier.</param>
+    /// <param name="orgClaimId">The organization claim identifier.</param>
```
SandGrainOne, line 191: The org claim id value is actually the organization number of the system owner. We could name the parameter something like that instead.

src/Events/Authorization/XacmlMapperHelper.cs
Ahmed-Ghanam: I renamed it.

SandGrainOne (approved):

Issue #627, opened 2024-11-29
Creating generic subscription returns 401 after failed authorization instead of 403


SandGrainOne:
### Description of the bug

Failed authorization when creating a generic subscription results in a 401 error where I would expect 403.

401 indicate something wrong with the identity and not authorization. We should return 403 if the error is because of access rights.

### Steps To Reproduce

1. Create a bearer token for a service owner
2. Attempt to create a subscription for events from a resource the service owner has no rights to
3. Observe that the result code is 401.

SandGrainOne:
Tested Create, Get and Delete as TTD with subscriber scopes. Works as expected.

PR #669, opened 2025-02-20
627 Should return Forbidden when subscribing to a resource without access

sigridbra:
## Description
Returning error code 403 Forbidden when the provided identity does not match the identity with access to a resource

## Related Issue(s)
- #627 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Events/Services/SubscriptionService.cs
+++ b/src/Events/Services/SubscriptionService.cs
@@ -48,7 +48,7 @@ public class SubscriptionService : ISubscriptionService
         if (!await _authorization.AuthorizeConsumerForEventsSubscription(eventsSubscription))
         {
             var errorMessage = $"Not authorized to create a subscription for resource {eventsSubscription.ResourceFilter}.";
-            return (null, new ServiceError(401, errorMessage));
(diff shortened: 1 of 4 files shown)


src/Events/Services/SubscriptionService.cs
```diff
@@ -72,7 +72,7 @@ public async Task<ServiceError> DeleteSubscription(int id)
 
         if (!AuthorizeAccessToSubscription(subscription))
```
sigridbra, line 73: Unsure about these as the issue only specifies when creating a subscription. This is for deleting a subscription. @SandGrainOne might clarify

src/Events/Services/SubscriptionService.cs
```diff
@@ -93,7 +93,7 @@ public async Task<ServiceError> DeleteSubscription(int id)
...
```
sigridbra, line 94: Unsure about these as the issue only specifies when creating a subscription. This is for getting a subscription. @SandGrainOne might clarify

SandGrainOne:
This is technically a breaking change, but I assume there are no systems that have explicit handling of 401 responses.

Issue #657, opened 2025-02-11
Create log service and repository layers


eskebab:
*Problem*
We need to have a design for communicating and testing database inserts

*Solution*
We use existing separation principles into a service layer that communicates with a repository layer. The service layer is injected into the api controller. The repository layer is injected into the service layer. The repository layer is concerned with all database communication. The service layer can only communicate with the database through the repository. This abstraction makes it easy to either mock or replace the repository implementation

- [x] Create two layers: service and repository
- [x] Repository class for database transactions. Will for now only support insert
- [x] Domain model class to represent a log entry
- [x] Service to communicate with repository class from controller
- [ ] Create unit tests for service
- [ ] Create integration tests for repository

PR #670, opened 2025-02-20
Feature/657 create layers

eskebab:
## Related Issue(s)
- #657 

## Verification
- [ ] All tests run green

## Documentation


src/Events/Services/TraceLogService.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            if (!parseResult)
+            {
+                _logger.LogError("Error creating trace log entry for webhook POST response: Invalid GUID");
+                return string.Empty;
+            }
+
+            if (string.IsNullOrEmpty(logEntryDto.CloudEventId) || logEntryDto.Endpoint == null || logEntryDto.StatusCode == null)
+            {
+                _logger.LogError("Error creating trace log entry for webhook POST response: Missing required input parameters");
```
hggutvik, line 91: Would it be an advantage to log which of (logEntryDto.CloudEventId, logEntryDto.Endpoint, logEntryDto.StatusCode) is null in this case?

src/Events/Services/Interfaces/ITraceLogService.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Net;
+using System.Threading.Tasks;
+using Altinn.Platform.Events.Models;
```
hggutvik, line 4: I'm not sure, but I think we have a practice of sorting and organzing usings, e.g. with Altinn packages separated in the top. Not a big deal, but it's nice for readability if we have some shared standard. @SandGrainOne , is alphabetical sorting a general thing in our codebases?

src/Events/Services/TraceLogService.cs
```diff
@@ -0,0 +1,117 @@
...
+                    SubscriptionId = logEntryDto.SubscriptionId,
+                    ResponseCode = (int?)logEntryDto.StatusCode,
+                    Activity = TraceLogActivity.WebhookPostResponse
+                };
+                await _traceLogRepository.CreateTraceLogEntry(traceLogEntry);
+                return logEntryDto.CloudEventId;
+            }
+            catch (Exception)
+            {
+                throw;
```
hggutvik: The other methods in this class are intentionally swallowing the exceptions from the repo. Should we do the same in here as well?

hggutvik (changes requested):
You've done a lot of good work here! A few small questions for you to look at, before I approve ; )

src/Events/Services/Interfaces/ITraceLogService.cs
eskebab: yes! good point

src/Events/Services/TraceLogService.cs
eskebab: Yes, I can add the parameters to the log statement

src/Events/Services/Interfaces/ITraceLogService.cs
SandGrainOne: In most cases you will find System namespaces at the top and I believe there is a rule for it somewhere. Namespaces should otherwise be sorted alphabetically. On top of that I like to group declarations by first word in namespace.

src/Events/Services/TraceLogService.cs
```diff
@@ -0,0 +1,110 @@
...
+                CloudEventId = parsedGuid,
+                Resource = logEntryDto.CloudEventResource,
+                EventType = logEntryDto.CloudEventType,
+                Consumer = logEntryDto.Consumer,
+                SubscriberEndpoint = logEntryDto.Endpoint.ToString(),
+                SubscriptionId = logEntryDto.SubscriptionId,
+                ResponseCode = (int?)logEntryDto.StatusCode,
+                Activity = TraceLogActivity.WebhookPostResponse
+            };
+            await _traceLogRepository.CreateTraceLogEntry(traceLogEntry);
```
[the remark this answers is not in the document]
eskebab: Yes, because it should be called by the storageController. I wanted to handle exceptions there, but I will describe this in the doc so it makes more sense

src/Events/Services/TraceLogService.cs
```diff
@@ -86,9 +92,15 @@ public async Task<string> CreateWebhookResponseEntry(LogEntryDto logEntryDto)
                 return string.Empty;
             }
 
-            if (string.IsNullOrEmpty(logEntryDto.CloudEventId) || logEntryDto.Endpoint == null || logEntryDto.StatusCode == null)
+            if (logEntryDto.Endpoint == null)
+            {
+                _logger.LogError("Error creating trace log entry for webhook POST response: Missing required Endpoint parameter");
+                return string.Empty;
```
hggutvik, line 98: Is there a chance that some of these error cases can occur in combination? I.e., invalid guid AND missing endpoint parameter? 

In that case, we could aggregate the invalid params into the shared log, e.g. by doing something like this:

```
public async Task<string> CreateWebhookResponseEntry(LogEntryDto logEntryDto)
{
    var parseResult = Guid.TryParse(logEntryDto.CloudEventId, out Guid parsedGuid);

    string errorMessage = Validate(parseResult, logEntryDto);

    if (string.IsNullOrEmpty(errorMessage))
    {
        var traceLogEntry = new TraceLog
        {
            CloudEventId = parsedGuid,
            Resource = logEntryDto.CloudEventResource,
            EventType = logEntryDto.CloudEventType,
            Consumer = logEntryDto.Consumer,
            SubscriberEndpoint = logEntryDto.Endpoint.ToString(),
            SubscriptionId = logEntryDto.SubscriptionId,
            ResponseCode = (int?)logEntryDto.StatusCode,
            Activity = TraceLogActivity.WebhookPostResponse
        };
        await _traceLogRepository.CreateTraceLogEntry(traceLogEntry);
        return logEntryDto.CloudEventId;
    }

    _logger.LogError("Error creating trace log entry for webhook POST response: {ErrorMessage}", errorMessage);
     return string.Empty;
    }
}

/// <summary>
/// Validates parameters
/// </summary>
/// <returns>A joined string of one or more errors, or string.Empty if params are valid </returns>
private static string Validate(bool parseResult, LogEntryDto logEntryDto)
{
    string[] errors = 
    [
        !parseResult ? "Invalid GUID" : string.Empty,
        logEntryDto.Endpoint == null ? "Missing required Endpoint parameter" : string.Empty,
        logEntryDto.StatusCode == null ? "Missing required StatusCode parameter" : string.Empty
    ];
    var aggregatedErrorMessage = string.Join(", ", errors.Where(s => !string.IsNullOrEmpty(s)));

    return aggregatedErrorMessage;
}
```
On a relevant (but nitpicking) sidenote concerning readability, it's nice to have a consistent structure of handling happypath vs errors. I don't know of any existing practice across the codebase, but in this class, the other methods have "first happypath, then error handling". If it's not too much of a hassle, it's nice to follow the same structure here (as in the code example of this comment).

hggutvik (approved):
I have one question that you might want to look at - if it's irrelevant, feel free to complete and merge 🚀

Issue #659, opened 2025-02-11
Cleanup routine using table partitioning


eskebab:
*Problem*
We want to keep data for as long as necessary to do proper analysis. When log entries are considered useless for troubleshooting purposes, they should be deleted or partitioned

*Solution*
We implement database table partitioning using built-in partitioning in postgres. We can use existing cron.schedule function IF we are able to get next month from context

eskebab:
Following these guidelines for creating a partitioned table:
https://www.postgresql.org/docs/current/ddl-partitioning.html

eskebab:
CREATE TABLE events.trace_log_y2024m03 PARTITION OF events.trace_log
    FOR VALUES FROM ('2025-03-01') TO ('2025-04-01');

creates a range partition where this insert will fall into partition:
_INSERT INTO events.trace_log(
	cloudeventid, resource, eventtype, consumer, "time", subscriptionid, responsecode, subscriberendpoint, activity)
	VALUES ('ec10976a-e507-4ef3-9ee8-1e44a52501be', 'resource', 'eventtype', 'consumer', '2025-03-01', 1, 200, 'http://localhost', 'TEST');_

while this entry will create an error, given that the above partition is the only available partition:
_INSERT INTO events.trace_log(
	cloudeventid, resource, eventtype, consumer, "time", subscriptionid, responsecode, subscriberendpoint, activity)
	VALUES ('ec10976a-e507-4ef3-9ee8-1e44a52501be', 'resource', 'eventtype', 'consumer', '2025-04-01', 1, 200, 'http://localhost', 'TEST');_

ERROR:  no partition of relation "trace_log" found for row
Partition key of the failing row contains ("time") = (2025-04-01 00:00:00+02).

eskebab:
If one tries to create a partition that overlaps with another, this error will be given:
_partition "trace_log_y2024m04" would overlap partition "trace_log_y2024m03"_

eskebab:
since creating a table partition is an idempotent operation, using a background service in .NET could be a good alternative, because we would then have control over the range of the coming month/partition.

eskebab:
Tried creating some partitions of various size. First, I created a 1month duration partition. Then I created one that has a duration of six months. I then created one with a duration of only 1 day.

eskebab:
funksjon pg_artition_tree() i Postgres for å hente ut partisjoner som tilhører en tabell:

select *
from  pg_partition_tree('events.trace_log'::regclass)
where parentrelid is not null;

PR #678, opened 2025-03-11
Add migration script for partitioning table trace_log

eskebab:
## Related Issue(s)
- #659 

## Verification
- [ ] All tests run green

## Documentation

Following these guidelines for creating a partitioned table:
https://www.postgresql.org/docs/current/ddl-partitioning.html

see comments in issue #659 regarding how this was tested


src/Events/Migration/v0.45/01-setup-tables.sql
```diff
@@ -0,0 +1,21 @@
+-- Table: events.trace_log
+
+DROP TABLE IF EXISTS events.trace_log;
+
+CREATE TABLE IF NOT EXISTS events.trace_log
+(
+    cloudeventid uuid NOT NULL,
+    resource character varying COLLATE pg_catalog."default",
```
HenningNormann, line 8: Why not data type text for all texts?

src/Events/Migration/v0.45/01-setup-tables.sql
```diff
@@ -0,0 +1,21 @@
...
+    subscriptionid bigint,
+    responsecode integer,
+    subscriberendpoint character varying COLLATE pg_catalog."default",
+    activity character varying COLLATE pg_catalog."default",
+    sequenceno BIGSERIAL,
+	CONSTRAINT trace_log_pkey PRIMARY KEY (sequenceno, time)
+    
+) PARTITION BY RANGE (time);
+
+CREATE INDEX ON events.trace_log (time);
```
HenningNormann, line 21: Shouldn't this index rather be created on the child tables?

eskebab: _Create an index on the key column(s), as well as any other indexes you might want, on the partitioned table. (The key index is not strictly necessary, but in most scenarios it is helpful.) This automatically creates a matching index on each partition, and any partitions you create or attach later will also have such an index._

I also tried testing this locally. It does indeed create the index on time for each individual partition.

src/Events/Migration/v0.45/01-setup-tables.sql
eskebab: events.subscription is defined using character varying for all text columns. 
For consistency? Up for discussion

SandGrainOne: Let's turn it around. Why text for all texts? 

The choice we made for subscription might not have been super well thought out. If text is better, then we use text.

src/Events/Migration/FunctionsAndProcedures/createlogspartition.sql
```diff
@@ -0,0 +1,44 @@
+CREATE OR REPLACE PROCEDURE events.deletesubscription(_id integer)
```
SandGrainOne, line 1: Something wrong here. The new function is defined inside a script for an existing function.

src/Events/Migration/v0.45/01-setup-tables.sql
SandGrainOne: @HenningNormann Does using `text` have an advantage?

src/Events/Migration/FunctionsAndProcedures/createlogspartition.sql
```diff
@@ -0,0 +1,37 @@
...
+   IF nextMonth = 13 THEN
+      nextMonth = 1;
+	  nextYear = nextYear + 1;
+  END IF;
+
+   fromMonth = thisYear || '-' || lpad(thisMonth::varchar, 2, '0') || '-01' ; 
+   toMonth = nextYear || '-' || lpad(nextMonth::varchar, 2, '0') || '-01';
+   partitionName = 'events.trace_log_y' || nextYear || 'm' || nextMonth;	
+   
+   EXECUTE
```
HenningNormann, line 32: No error handling?

src/Events/Migration/v0.45/01-setup-tables.sql
eskebab: Changed all character varying columns to text and ran integration test with API locally. No changes in repository class necessary based on type change.

src/Events/Migration/FunctionsAndProcedures/createlogspartition.sql
eskebab: I was thinking better to let the error bubble up and handle it elsewhere, but I'm open to suggestions. Most importantly, creating a new partition several times does not result in duplicate partitions.

SandGrainOne (approved):

PR #697, opened 2025-03-28
Convert national identity number to party UUID for authorization

SandGrainOne:
## Description
The PEP package we're using to simplify authorization doesn't accept resource element that contains national identity numbers. It's causing a syntax error. In this change we're replacing the nin with the party uuid through a register lookup before calling Authorization.

- Single out all unique event subjects of the type we need to replace
- Perform call to register that can provide the correct party uuid values for each person
    - Register has a limit of 100 so we're chunking the request based on that.
    - Combine the responses into one larger list
- Loop through the events with nin in subjects
    - Take a backup of the original event subject and store it in a temporary attribute within the event
    - Overwrite event subject with a party uuid identifier where we found a match in register, otherwise set to null
- Perform authorization
   - Loop through authorized events and revert the temporary manipulation of subject for all events.

## Related Issue(s)
- #653 

## Verification
- [x] All tests run green

## Documentation


src/Events/Services/EventsService.cs
```diff
@@ -19,13 +21,15 @@ namespace Altinn.Platform.Events.Services
...
+        private const string _originalSubjectKey = "originalsubjectreplacedforauthorization";
+
         private readonly ICloudEventRepository _repository;
         private readonly ITraceLogService _traceLogService;
         private readonly IEventsQueueClient _queueClient;
 
         private readonly IRegisterService _registerService;
         private readonly IAuthorization _authorizationService;
-        private readonly ILogger<IEventsService> _logger;
+        private readonly ILogger _logger;
```
eskebab, line 32: why not ILogger<EventsService>?

eskebab (approved):

PR #972, opened 2026-03-27
Feature/967 include main unit subscriptions

### LINKED ISSUE #967: Add subscription search for main units
## Description
The subscription system in Events are currently very restrictive when it comes to subject filtering. The subject filter must match the event subject exactly. There are no lookup to check for main to sub-unit hierarchy that could allow a subscription defined on the main unit to catch sub-unit related events.

This issue will try to remedy this by checking if the subject of an event has a parent unit and then search for matching subscriptions on that in addition to direct sub-unit matching.

In order to prevent an explosion of events to existing main unit subscriptions we should consider introducing a subscription option where this extra lookup would need to be activated. `IncludeSubUnitEvents` or something similar.

There should be caching of the sub-unit to main-unit lookup result to reduce the number of calls against Register. It should be safe to cache this for many hours. These relationships doesn't change that often. Register has a specialized endpoint for our needs that we're already using in Profile for similar cases.

Events would need to search for possible subscriptions using both units. This would require us to either perform two searches or updating the call chain to support an array of subjects. The last option is preferred. This could also help resolve the existing issue with subscription subject matching documented below.

## Solution
- Added main unit lookup for when subject is an organization (or unknown type)
- Added a new field called IncludeSubunits to subscriptions to enable opt-in mechanism around main unit subscription search.
- Added search for main unit subscriptions with IncludeSubunits turned on when subject is a subunit.

IncludeSubunit is a new unique property of Subscriptions during subscription creation.
 - Subscriptions where IncludeSubunits are NULL or FALSE are different from subscriptions where IncludeSubunits are true.
 - Subscriptions where IncludeSubunits are NULL are equal to subscriptions where IncludeSubunits are false.

SandGrainOne:
## Description
Add a feature that will allow a main unit subscription to also catch events generated for a subunit.

- Added a IncludeSubunit flag to subscription in order to make the feature opt in per subscription.
- Expanded RegisterService with a method to retrieve main unit for sub units.
- Added an extra subscription search for main unit subscriptions if an event is for a sub unit.

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Added IncludeSubunits subscription flag and main‑unit resolution to optionally include main‑unit subscriptions in outbound delivery; main‑unit lookup is cached.

* **Bug Fixes**
  * Validation rejects IncludeSubunits when no subject filter is provided; outbound merging of main‑unit subscriptions only occurs when appropriate.

* **Tests**
  * Added unit, integration, and HTTP tests covering main‑unit lookup, alternative‑subject handling, caching, and subunit inclusion.

* **Chores**
  * Database migration, seed data and API updates to persist and expose the new flag.


src/Events/Migration/v0.52/02-functions-and-procedures.sql
```diff
@@ -0,0 +1,282 @@
...
+    FROM
+        events.subscription s
+    WHERE
+        s.resourcefilter = _resourcefilter
+        AND ((_sourcefilter IS NULL AND s.sourcefilter IS NULL) OR s.sourcefilter = _sourcefilter)
+        AND ((_subjectfilter IS NULL AND s.subjectfilter IS NULL) OR s.subjectfilter = _subjectfilter)
+        AND ((_typefilter IS NULL AND s.typefilter IS NULL) OR s.typefilter = _typefilter)
+        AND s.consumer = _consumer
+        AND s.endpointurl = _endpointurl
+        AND ((_includesubunits IS NULL AND s.includesubunits IS NULL) OR s.includesubunits = _includesubunits);
```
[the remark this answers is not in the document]
SandGrainOne: Generated file. Content copied from files in the FunctionsAndProcedures folder.

src/Events/Migration/v0.52/02-functions-and-procedures.sql
```diff
@@ -0,0 +1,282 @@
...
+	FROM events.events
+	WHERE cloudevent->>'id' = _after;
+END IF;
+SELECT MAX(sequenceno) INTO _sequenceno_last FROM events.events WHERE registeredtime <= now() - interval '30 second';
+return query
+	SELECT cast(cloudevent as text) as cloudevents
+		FROM events.events
+		WHERE (_subject IS NULL OR cloudevent->>'subject' = _subject)
+			AND (_from IS NULL OR cloudevent->>'time' >= to_json(_from)::text)
+			AND (_to IS NULL OR cloudevent->>'time' <= to_json(_to)::text)
```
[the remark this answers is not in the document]
SandGrainOne: Generated file. Content copied from files in the FunctionsAndProcedures folder.

src/Events/Migration/v0.52/02-functions-and-procedures.sql
```diff
@@ -0,0 +1,282 @@
...
+AS $BODY$
+DECLARE currentTime timestamptz;
+
+BEGIN
+	SET TIME ZONE UTC;
+	currentTime := NOW();
+
+	RETURN QUERY
+	INSERT INTO events.subscription(resourcefilter, sourcefilter, subjectfilter, typefilter, consumer, endpointurl, createdby, "time", validated, includesubunits)
+	VALUES ($1, $2, $3, $4, $5, $6, $7, currentTime, $8, $9) RETURNING *;
```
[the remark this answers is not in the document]
SandGrainOne: Generated file. Content copied from files in the FunctionsAndProcedures folder.

src/Events/Migration/v0.52/02-functions-and-procedures.sql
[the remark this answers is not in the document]
SandGrainOne: When making changes like this it's important that there are appropriate indexes. Is that a concern here?

src/Events/Services/RegisterService.cs
```diff
@@ -132,5 +132,39 @@ await response.Content.ReadFromJsonAsync<PartiesRegisterQueryResponse>(
...
+            if (!response.IsSuccessStatusCode)
+            {
+                throw await PlatformHttpException.CreateAsync(response);
+            }
+
+            var responseObject = await response.Content.ReadFromJsonAsync<LookupMainUnitResponse>(_serializerOptions, cancellationToken);
+            if (!(responseObject?.Data?.Count > 0))
+            {
+                return null;
+            }
```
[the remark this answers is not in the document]
SandGrainOne: This code is outside a nullable annotation context. This will be adressed at some point, but not now.

src/Events/Services/GenericSubscriptionService.cs
```diff
@@ -66,6 +66,12 @@ private static bool ValidateSubscription(Subscription eventsSubscription, out st
             return false;
         }
 
+        if (eventsSubscription.IncludeSubunits != null && string.IsNullOrEmpty(eventsSubscription.SubjectFilter))
```
sigridbra:
```suggestion
        if (eventsSubscription.IncludeSubunits == true && string.IsNullOrEmpty(eventsSubscription.SubjectFilter))
```

src/Events/Services/AppSubscriptionService.cs
```diff
@@ -112,6 +112,12 @@ private bool ValidateSubscription(Subscription eventsSubscription, out string me
                 return false;
             }
 
+            if (eventsSubscription.IncludeSubunits != null && string.IsNullOrEmpty(eventsSubscription.SubjectFilter))
```
sigridbra: I assume IncludeSubunits = null and false should be treated equally. 
```suggestion
            if (eventsSubscription.IncludeSubunits == true && string.IsNullOrEmpty(eventsSubscription.SubjectFilter))
```

src/Events/Models/Subscription.cs
```diff
@@ -67,6 +67,11 @@ public class Subscription
         /// </summary>
         public bool Validated { get; set; }
 
+        /// <summary>
+        /// Indicates whether events for sub-units of the subject should be included.
+        /// </summary>
+        public bool? IncludeSubunits { get; set; }
```
sigridbra: Why do we need this to be nullable? If this is for DB performance it should not be a problem to add a column with default false. 

> Adding a column with a constant default value does not require each row of the table to be updated when the ALTER TABLE statement is executed
Source: [https://www.postgresql.org/docs/current/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN](https://www.postgresql.org/docs/current/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN)

src/Events/Services/OutboundService.cs
```diff
@@ -79,9 +84,77 @@ public async Task PostOutbound(CloudEvent cloudEvent, CancellationToken cancella
              cloudEvent.Type,
              cancellationToken);
 
+        string? alternativeSubject = cloudEvent["alternativesubject"]?.ToString();
+        var (mainUnitLookupUrn, originalFormat) = GetSubjectLookupDetails(cloudEvent.Subject, alternativeSubject);
```
sigridbra, line 88: It feels a bit backwards to lookup the main unit even if there might be zero subscriptions for the main unit with IncludeSubunits = true, but I cannot see a better way to handle this right now.

src/Events/Models/Subscription.cs
SandGrainOne: I didn't know. Didn't check. I'll see if I can change it.

src/Events/Services/OutboundService.cs
SandGrainOne: Same, same.

src/Events/Models/SubscriptionRequestModel.cs
```diff
@@ -41,6 +41,11 @@ public class SubscriptionRequestModel
         /// </summary>
         public string TypeFilter { get; set; }
 
+        /// <summary>
+        /// Indicates whether events for sub-units of the subject should be included.
+        /// </summary>
+        public bool IncludeSubunits { get; set; }
```
sigridbra: Will this make the analysis happy? 
```suggestion
        public bool IncludeSubunits { get; set; } = false;
```

sigridbra (approved):

sigridbra (approved):
Looks good!

PR #991, opened 2026-05-08
Pass SONAR_TOKEN to dotnet-sonarscanner via existing env var

hggutvik:
We have an existing and unused env var for SONAR_TOKEN in the Sonar script steps.
We should either use the env var or remove it and stick to the inline `${{ secrets.SONAR_TOKEN }}`.

For secrets specifically, GitHub masks the value in logs regardless of whether you use an env var or inline ${{ secrets.* }}, so the practical security difference is negligible either way.

I have chosen using the env var instead of inline ${{ }} as this is the general recommendation from GitHub in how to handle values that could be user-controlled (e.g. PR titles, issue bodies). For secrets, that risk doesn't apply since secrets aren't attacker-controlled. 
Consistently using env vars for secrets — even when there's no practical difference — reinforces the habit that matters when you do have user-controlled input. Consistency also makes code easier to review, since a reviewer doesn't have to reason about whether a particular inline ${{ }} is safe or not.

GitHub doc ref: https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

## Summary by CodeRabbit

* **Chores**
  * Updated build and analysis workflow configuration to optimize environment variable handling and trigger conditions.

---

**Note:** This release contains internal infrastructure updates with no user-facing changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/altinn-events/pull/991)

### DIFF
--- a/.github/workflows/build-and-analyze.yml
+++ b/.github/workflows/build-and-analyze.yml
@@ -1,9 +1,9 @@
 name: Code test and analysis
 on:
   push:
-    branches: [ main ]
+    branches: [main]
     paths:
-    - 'src/**'
+      - 'src/**'
   pull_request:
     types: [opened, synchronize, reopened]
     paths:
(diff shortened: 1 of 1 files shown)


hggutvik:
@coderabbitai review

sigridbra (approved):

PR #1043, opened 2026-06-29
Fix scoped wolverine service

martivj:
## Description

Wolverine v6 changes the default `ServiceLocationPolicy` from `AllowedButWarn` to `NotAllowed`. Under the old default, Wolverine would fall back to resolving services at runtime (via `GetService`) when it couldn't statically trace a service's dependency graph — logging a warning but continuing. With `NotAllowed`, those same cases now throw an `InvalidOperationException` at the point a handler is first invoked, when Wolverine attempts to generate the execution code for it.

Two registration patterns in the codebase were incompatible with static tracing:

- **Typed HTTP clients** (`AddHttpClient<IInterface, TImpl>()`) register the service via an opaque factory delegate. Wolverine cannot inspect `ImplementationType` on the resulting `ServiceDescriptor`, so it can't generate handler code for any service that depends on it.
- **Scrutor's `Decorate<>()`** always registers the decorator through a lambda factory regardless of lifetime, with the same result.

The fix addresses root causes rather than overriding `ServiceLocationPolicy` back to `AllowedButWarn`. Restoring the old default would silence the errors but reintroduce silent runtime service location — bypassing Wolverine's static code generation, losing startup-time dependency validation, and making the handler execution path slower and less predictable. Fixing the registrations properly means Wolverine generates efficient static handler code and will fail fast if the dependency graph ever becomes unresolvable again.

Concretely:
- Typed HTTP clients replaced with named clients (`AddHttpClient(name)`) paired with explicit `AddScoped<>()` — giving Wolverine a clean `ServiceDescriptor` with a known `ImplementationType`
- Scrutor decoration replaced with explicit registrations via a new `IRegisterApiClient` interface, which `RegisterApiClient` implements and `CachingRegisterService` depends on — letting Wolverine trace the full chain without opaque factories

The rename from `RegisterService`/`RegisterServiceCachingDecorator` to `RegisterApiClient`/`CachingRegisterService` was done at the same time to make each type's role unambiguous.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook service initialization and HTTP client handling for more reliable outbound requests.
  * Kept existing timeout behavior and request logic unchanged.

### DIFF
--- a/src/Events/Services/RegisterApiClient.cs
+++ b/src/Events/Services/RegisterApiClient.cs
@@ -43,7 +43,7 @@ namespace Altinn.Platform.Events.Services
         private const string ApplicationJson = "application/json";
 
         /// <summary>Name of the named <see cref="HttpClient"/> used by this service.</summary>
-        internal const string _httpClientName = nameof(RegisterApiClient);
+        internal const string HttpClientName = nameof(RegisterApiClient);
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RegisterApiClient"/> class.
@@ -56,7 +56,7 @@ namespace Altinn.Platform.Events.Services
             IOptions<PlatformSettings> platformSettings,
             ILogger<RegisterApiClient> logger)
         {
-            _client = httpClientFactory.CreateClient(_httpClientName);
+            _client = httpClientFactory.CreateClient(HttpClientName);
             _httpContextAccessor = httpContextAccessor;
             _generalSettings = generalSettings.Value;
             _accessTokenGenerator = accessTokenGenerator;
--- a/src/Events/Services/RegisterApiClient.cs
+++ b/src/Events/Services/RegisterApiClient.cs
@@ -40,6 +40,8 @@ namespace Altinn.Platform.Events.Services
 
         private readonly int _chunkSize;
 
+        private const string ApplicationJson = "application/json";
+
         /// <summary>Name of the named <see cref="HttpClient"/> used by this service.</summary>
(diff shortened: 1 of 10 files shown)


src/Events/Services/RegisterApiClient.cs
```diff
@@ -40,25 +40,30 @@ public class RegisterService : IRegisterService
 
         private readonly int _chunkSize;
 
+        private const string ApplicationJson = "application/json";
+
+        /// <summary>Name of the named <see cref="HttpClient"/> used by this service.</summary>
+        internal const string _httpClientName = nameof(RegisterApiClient);
```
SandGrainOne:
```suggestion
        internal const string HttpClientName = nameof(RegisterApiClient);
```

src/Events/Services/WebhookService.cs
```diff
@@ -23,13 +23,16 @@ public class WebhookService : IWebhookService
         private readonly ILogger<WebhookService> _logger;
         private readonly string _slackUri = "hooks.slack.com";
 
+        /// <summary>Name of the named <see cref="HttpClient"/> used by this service.</summary>
+        internal const string _httpClientName = nameof(WebhookService);
```
SandGrainOne:
```suggestion
        internal const string HttpClientName = nameof(WebhookService);
```

SandGrainOne (approved):

PR #1052, opened 2026-07-10
Bug/977 validate source filter scheme to be https

### LINKED ISSUE #977: Incomplete source filter validation allows for non-functional subscriptions
### Description of the bug

It's currently possible to register subscriptions where source filter is outside the possible values for an event source:

1. There are subscriptions where source filter starts with `http://` instead of `https://`
    e.g: http://nb.apps.altinn.no/nb/deponering-ekspropriasjon
3. There are subscriptions where source filter looks for app events from platform
    e.g: https://platform.altinn.no/staf/ressursrapportering

The subscriptions are currently working because we aren't using the source filter directly. We used the path to generate a resource filter which is used instead. This logic is apparently not to picky about the input.

In issue #930 we're currently working on a fix where source filter again will be respected. This will effectively disable subscriptions with these errors. We'll need to contact "nb" and "staf" and inform them of the changes. We can update the existing subscriptions, but they'll start getting errors when trying to create new subscriptions.

### Steps To Reproduce

Following subscriptions should currently go through if the request identity is the service owner.
Try request:
``` json
{
  "sourceFilter": "http://ttd.apps.at23.altinn.cloud/ttd/apps-test",
  "endpoint": "https://webhook.site/73e46e5f-e905-4375-8ab6-12986525007e",
}
```

Try request:
``` json
{
  "sourceFilter": "https://platform.apps.at23.altinn.cloud/ttd/apps-test",
  "endpoint": "https://webhook.site/73e46e5f-e905-4375-8ab6-12986525007e",
}
```

## Analysis
Full validation of source filter is prone to break. There are too many conditions related to what a valid source would look like. We'll keep the check to ensure that the string is a valid uri and add a check to ensure the schema is https. No other changes for now. 

Systems that generate subscriptions will hopefully notice if they're not getting any events. We also prefer if subscriptions didn't have a source filter at all, but relied on resource filter instead. The only place were this could result in too many events is for apps where there is a secondary source in the case of Dialogporten.

SandGrainOne:
## Description
Improve validation of both the source filter and endpoint addresses of a new subscription.

Both needs to be an absolute uri and the echeme must be https. EndPoint is required while source filter is optional.

## Related Issue(s)
- #977 

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscription endpoints now require absolute HTTPS URLs.
  * Source filters now require absolute HTTPS URLs when provided.
  * Invalid resource filter values return clearer, exact validation messages.
  * Improved validation ordering for clearer 400 responses.

* **Tests**
  * Expanded negative validation coverage for subscriptions (endpoints, source filters, and resource filters).
  * Renamed and adjusted tests to match the stricter HTTPS URL rules.

### DIFF
--- a/src/Events/Controllers/SubscriptionController.cs
+++ b/src/Events/Controllers/SubscriptionController.cs
@@ -61,12 +61,12 @@ namespace Altinn.Platform.Events.Controllers
         [Produces("application/json")]
         public async Task<ActionResult<Subscription>> Post([FromBody] SubscriptionRequestModel subscriptionRequest)
         {
-            if (subscriptionRequest.EndPoint == null || !(subscriptionRequest.EndPoint.IsAbsoluteUri && subscriptionRequest.EndPoint.Scheme == Uri.UriSchemeHttps))
+            if (!IsValidUri(subscriptionRequest.EndPoint))
             {
                 return StatusCode(400, "Missing or invalid endpoint to push events towards");
             }
 
-            if (subscriptionRequest.SourceFilter != null && !(subscriptionRequest.SourceFilter.IsAbsoluteUri && subscriptionRequest.SourceFilter.Scheme == Uri.UriSchemeHttps))
+            if (subscriptionRequest.SourceFilter is not null && !IsValidUri(subscriptionRequest.SourceFilter))
             {
                 return StatusCode(400, "SourceFilter must be an absolute URL with the https scheme.");
             }
@@ -188,6 +188,16 @@ namespace Altinn.Platform.Events.Controllers
             return Ok();
         }
 
+        private static bool IsValidUri(Uri uri)
+        {
+            return uri is not null && uri.IsAbsoluteUri && uri.Scheme == Uri.UriSchemeHttps;
+        }
+
+        private static void AddQueryModelToTelemetry(SubscriptionRequestModel subscriptionRequest)
+        {
+            Activity.Current?.AddTag("subscription.request", JsonSerializer.Serialize(subscriptionRequest));
+        }
+
         private bool IsAppSubscription(SubscriptionRequestModel subscription)
         {
             if (subscription.ResourceFilter != null &&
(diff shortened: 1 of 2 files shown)


src/Events/Controllers/SubscriptionController.cs
```diff
@@ -61,9 +61,14 @@ public SubscriptionController(
         [Produces("application/json")]
         public async Task<ActionResult<Subscription>> Post([FromBody] SubscriptionRequestModel subscriptionRequest)
         {
-            if (subscriptionRequest.SourceFilter != null && !Uri.IsWellFormedUriString(subscriptionRequest.SourceFilter.ToString(), UriKind.Absolute))
```
hggutvik, line 64: Noticing we had `Uri.IsWellFormedUriString` before (although on ToString rather than OriginalString, where apparently .NET can falsely gives us a cleaned-up version in ToString), [I added some tests on another branch](https://github.com/Altinn/altinn-events/compare/bug/977-validate-source-filter-scheme-to-be-https...bug/977-keep-wellformed-uri-check?expand=1#diff-aaa887602c7eb6785f7a4c03ce0964e67b69cc0495ecc521588a24209dc75012) to see if we now lack RFC well-formedness / escaping validation. It seems that these Uri's won't be rejected with 400 in the current implementation.

From what I understand, our validation is meant to reject these. If so, we might want to add `Uri.IsWellFormedUriString(uri.OriginalString, UriKind.Absolute)` as a validation condition. 
[In my experimenation branch](https://github.com/Altinn/altinn-events/compare/bug/977-validate-source-filter-scheme-to-be-https...bug/977-keep-wellformed-uri-check), that makes the tests pass

SandGrainOne: I didn't look at this too deeply to be honest. I just thought it was an unnecessary step to stringify what was already successfully turned into a Uri through model binding. 

SubscriptionRepository takes the AbsoluteUri when storing the value to the database. I'm not too worried about unescaped special characters in weird places. They are handled.

hggutvik (approved):

PR #1054, opened 2026-07-10
Coderabbit should ignore migration version folders

SandGrainOne:
## Description
Trying to have coderabbit look at files outside the src/Events/Migration/FunctionsAndProcedures folder. Just not the version folders.

## Related Issue(s)
- Reduce some unnecessary coderabbit work on generated files.

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,4 +1,3 @@
 reviews:
   path_filters:
-    - "!src/Events/Migration/**"
-    - "src/Events/Migration/FunctionsAndProcedures/**"
+    - "!src/Events/Migration/v*/**"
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -1,4 +1,3 @@
 reviews:
   path_filters:
-    - "!src/Events/Migration/**"
-    - "src/Events/Migration/FunctionsAndProcedures/**"
+    - "!src/Events/Migration/v*/**"
```
sigridbra: Would it be possible to do something like 
    - "!src/Events/Migration/v*/*functions-and-procedures.sql"
to be able to include analysis on tables/indexes/etc, while skipping review on autogenerated file?

SandGrainOne: yes, probably. I can change right away. I got the same idea right after making the PR, but had already started on something else.

martivj (approved):

PR #1056, opened 2026-07-15
Minor improvements for labeler

hggutvik:
Remove redundant checkout-step.
Try and keep the labeler active as new commits are pushed to already opened PR (currenty not working..)


martivj (changes requested):
Revert the temporary

